PR #98, opened 2018-07-11
readme had unnecessary work

ranman:
https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/docker_native_scripts/amazonlinux-docker/Dockerfile#L75-L77
makes setting the environment variables irrelevant

*Issue #, if available:*

*Description of changes:*


ranman:
I don't expect this to be merged just wanted to point out some errors and figured this diff would be the easiest way.

unicornss:
Hi @ranman 

Thanks for the PR. 
We will revise those changes to the Dockerfile along with the change for installing gcc using
 `yum install gcc72-c++.x86_64` (instead of compiling it from sources).

Thanks.

PR #129, opened 2018-09-17
raspberry pi gst-launch-1.0 option misspell

JeongtaekLim:
*Description of changes:*

Option `periodicity-idr` doen't work for me and `periodicty-idr` worked.
Actually I'm not sure about it whether it is misspell or problem of my running environment.

ps. I don't expect this to be merged like #98  
But just want to point out options and help other people as possible I can!
Thank you for keep this library.


chehefen:
This has been fixed by new releases

PR #191, opened 2019-03-11
Aligning parameter ordering

gyorgy-gosi-epam:
*Description of changes:*
Aligning parameter ordering.

As it is in the README file: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/docker_native_scripts/README.md
Which is the proper way, also the AWS SDK example page is misaligned: https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/examples-rtsp.html


zhiyua-git:
Thanks for the catch. @gyorgy-gosi-epam  We have updated the docker image and [readme](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/docker_native_scripts/README.md) to keep aligned with our AWS document.

PR #316, opened 2019-12-27
Use pthread_cond_timedwait monotonic clock

ek9852:
Windows does not use pthread, but
in Linux or MACOS, pthread condition wait
use realtime clock which is affected by
ntp adjustment and should be avoid as possible,
use monotonic timer (linux)  and relative_np (macos)
to avoid that.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/kinesis-video-pic/src/utils/src/Mutex.c
+++ b/kinesis-video-pic/src/utils/src/Mutex.c
@@ -170,10 +170,6 @@ CleanUp:
 
 #else
 
-// Definition of the static global mutexes
-pthread_mutex_t globalKvsReentrantMutex = GLOBAL_MUTEX_INIT_RECURSIVE;
-pthread_mutex_t globalKvsNonReentrantMutex = GLOBAL_MUTEX_INIT;
-
 MUTEX defaultCreateMutex(BOOL reentrant)
 {
     pthread_mutex_t* pMutex;
@@ -181,18 +177,10 @@ MUTEX defaultCreateMutex(BOOL reentrant)
 
     // Allocate the mutex
     pMutex = (pthread_mutex_t*) MEMCALLOC(1, SIZEOF(pthread_mutex_t));
-    if (NULL == pMutex) {
-        return (MUTEX) (reentrant ? &globalKvsReentrantMutex : &globalKvsNonReentrantMutex);
-    }
 
-    if (0 != pthread_mutexattr_init(&mutexAttributes) ||
-        0 != pthread_mutexattr_settype(&mutexAttributes, reentrant ? PTHREAD_MUTEX_RECURSIVE : PTHREAD_MUTEX_NORMAL) ||
-        0 != pthread_mutex_init(pMutex, &mutexAttributes))
-    {
-        // In case of an error return the global mutexes
-        MEMFREE(pMutex);
-        return (MUTEX) (reentrant ? &globalKvsReentrantMutex : &globalKvsNonReentrantMutex);
-    }
+    pthread_mutexattr_init(&mutexAttributes);
+    pthread_mutexattr_settype(&mutexAttributes, reentrant ? PTHREAD_MUTEX_RECURSIVE : PTHREAD_MUTEX_NORMAL);
+    pthread_mutex_init(pMutex, &mutexAttributes);
 
     return (MUTEX) pMutex;
 }
@@ -217,24 +205,21 @@ VOID defaultFreeMutex(MUTEX mutex)
     pthread_mutex_t* pMutex = (pthread_mutex_t*) mutex;
     pthread_mutex_destroy(pMutex);
 
-    // De-allocate the memory if it's not a well-known mutex - aka if we had allocated it previously
-    if (pMutex != &globalKvsReentrantMutex && pMutex != &globalKvsNonReentrantMutex) {
-        MEMFREE(pMutex);
-    }
+    MEMFREE(pMutex);
 }
-
-pthread_cond_t globalKvsConditionVariable = PTHREAD_COND_INITIALIZER;
 
 CVAR defaultConditionVariableCreate()
 {
     CVAR pVar = (CVAR)MEMCALLOC(1, SIZEOF(pthread_cond_t));
-    if (NULL == pVar) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_t attr;
 
-    if (0 != pthread_cond_init(pVar, NULL)) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_init (&attr);
+#if defined __linux__ || defined __CYGWIN__
+    pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
+#endif
+
+    pthread_cond_init(pVar, NULL);
+    pthread_condattr_destroy (&attr);
 
     return pVar;
 }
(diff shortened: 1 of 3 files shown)


kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -170,29 +170,17 @@ STATUS defaultConditionVariableWait(CVAR cvar, MUTEX mutex, UINT64 timeout)
 
 #else
 
-// Definition of the static global mutexes
```
MushMal: Why are you removing these?

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -170,29 +170,17 @@ STATUS defaultConditionVariableWait(CVAR cvar, MUTEX mutex, UINT64 timeout)
...
 
-    if (0 != pthread_mutexattr_init(&mutexAttributes) ||
-        0 != pthread_mutexattr_settype(&mutexAttributes, reentrant ? PTHREAD_MUTEX_RECURSIVE : PTHREAD_MUTEX_NORMAL) ||
-        0 != pthread_mutex_init(pMutex, &mutexAttributes))
-    {
-        // In case of an error return the global mutexes
-        MEMFREE(pMutex);
-        return (MUTEX) (reentrant ? &globalKvsReentrantMutex : &globalKvsNonReentrantMutex);
-    }
+    pthread_mutexattr_init(&mutexAttributes);
```
MushMal: We should keep the result checking here and everywhere else

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -217,24 +205,21 @@ VOID defaultFreeMutex(MUTEX mutex)
     pthread_mutex_t* pMutex = (pthread_mutex_t*) mutex;
     pthread_mutex_destroy(pMutex);
 
-    // De-allocate the memory if it's not a well-known mutex - aka if we had allocated it previously
-    if (pMutex != &globalKvsReentrantMutex && pMutex != &globalKvsNonReentrantMutex) {
-        MEMFREE(pMutex);
-    }
+    MEMFREE(pMutex);
```
MushMal: same

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -217,24 +205,21 @@ VOID defaultFreeMutex(MUTEX mutex)
...
-    if (NULL == pVar) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_t attr;
 
-    if (0 != pthread_cond_init(pVar, NULL)) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_init (&attr);
+#if defined __linux__ || defined __CYGWIN__
```
MushMal: Do we need this condition? This seems to be a portable API

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -217,24 +205,21 @@ VOID defaultFreeMutex(MUTEX mutex)
...
     CVAR pVar = (CVAR)MEMCALLOC(1, SIZEOF(pthread_cond_t));
-    if (NULL == pVar) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_t attr;
 
-    if (0 != pthread_cond_init(pVar, NULL)) {
-        return &globalKvsConditionVariable;
-    }
+    pthread_condattr_init (&attr);
```
MushMal: Same - need to keep all of the result validation checks

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -284,7 +267,12 @@ STATUS defaultConditionVariableWait(CVAR cvar, MUTEX mutex, UINT64 timeout)
     INT32 retVal = 0;
     struct timespec timeSpec;
     pthread_mutex_t* pMutex = (pthread_mutex_t*) mutex;
+
+#if defined __linux__ || defined __CYGWIN__ || defined __MACH__
```
MushMal: Same question - perhaps we should have the same behavior for all of the platforms.

kinesis-video-pic/src/utils/src/Mutex.c
```diff
@@ -294,9 +282,15 @@ STATUS defaultConditionVariableWait(CVAR cvar, MUTEX mutex, UINT64 timeout)
         CHK(0 == (retVal = pthread_cond_wait(cvar, pMutex)), STATUS_WAIT_FAILED);
     }
     else {
+#ifdef __MACH__
```
MushMal: Is there an advantage using non-portable API for mac?

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: We should not shared a  global mutexes  if mutex creation was failed. It should assert instead.

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: agree

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: don't recommend  using a global mutex  when mutex creation failed. It can dead lock or create undesired effects, since the application assume every time a  mutex created is different., not a shared one.

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: macosx does not like it.

kinesis-video-pic/src/utils/src/Mutex.c
MushMal: Agree. We are not using these anywhere and none of the customers (in my opinion) need these.

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: It is same, since windows does not support pthread natively will  not run this portion of code

kinesis-video-pic/src/utils/src/Mutex.c
MushMal: We don't even need to assert on a failure - simply clean up and return INVALID_MUTEX_HANDLE_VALUE

kinesis-video-pic/src/utils/src/Mutex.c
ek9852: I am using a MBP with macosx, sure want this to be supported also.

kinesis-video-pic/src/utils/src/Mutex.c
MushMal: Indeed it doesn't

kinesis-video-pic/src/utils/src/Mutex.c
MushMal: Yeah, windows code is in the above section LN 115

kinesis-video-pic/src/utils/src/Mutex.c
MushMal: MBP? = Mack Book Pro? I believe the existing code should support it - I also use macbook pro as one of my development devices

ek9852: yep MBP = MacBookPro, posix pthread does not defined/support monotonic time.
gnu gcc added the extension, and apple added its own extension.
therefore we call different api call based on build environment.

MushMal:
I am going to resolve/close this issue. We have implemented a different clock macro - GETREALTIME()  in PIC that only controls the condition variable timeouts.

PR #337, opened 2020-02-08
Use absolute timestamp as default in gstreamer samples/plugin

zhiyua-git:
*Issue #, if available:*
N/A

*Description of changes:*
Use absolute timestamp as default in gstreamer samples/plugin

### DIFF
--- a/kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
+++ b/kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
@@ -42,7 +42,7 @@ LOGGER_TAG("com.amazonaws.kinesis.video.gstreamer");
 #define DEFAULT_TIMECODE_SCALE_MILLISECONDS 1
 #define DEFAULT_KEY_FRAME_FRAGMENTATION TRUE
 #define DEFAULT_FRAME_TIMECODES TRUE
-#define DEFAULT_ABSOLUTE_FRAGMENT_TIMES FALSE
+#define DEFAULT_ABSOLUTE_FRAGMENT_TIMES TRUE
 #define DEFAULT_FRAGMENT_ACKS TRUE
 #define DEFAULT_RESTART_ON_ERROR TRUE
 #define DEFAULT_RECALCULATE_METRICS TRUE
@@ -87,7 +87,9 @@ typedef struct _CustomData {
             last_unpersisted_file_idx(0),
             kinesis_video_producer(nullptr),
             kinesis_video_stream(nullptr),
-            main_loop(NULL) {
+            main_loop(NULL),
+            first_pts(GST_CLOCK_TIME_NONE),
+            use_absolute_fragment_times(true) {
         producer_start_time = chrono::duration_cast<nanoseconds>(systemCurrentTime().time_since_epoch()).count();
     }
 
@@ -157,6 +159,11 @@ typedef struct _CustomData {
     unique_ptr<Credentials> credential;
 
     uint32_t total_track_count;
+
+    bool use_absolute_fragment_times;
+
+    // Pts of first frame
+    uint64_t first_pts;
 
     GstElement *pipeline;
 } CustomData;
@@ -470,6 +477,14 @@ static GstFlowReturn on_new_sample(GstElement *sink, CustomData *data) {
         if (CHECK_FRAME_FLAG_KEY_FRAME(kinesis_video_flags)) {
             data->key_frame_pts = buffer->pts;
         }
+    } else if (data->use_absolute_fragment_times) {
+        if (data->first_pts == GST_CLOCK_TIME_NONE) {
+            data->first_pts = buffer->pts;
+        }
+        if (data->producer_start_time == GST_CLOCK_TIME_NONE) {
+            data->producer_start_time = chrono::duration_cast<nanoseconds>(systemCurrentTime().time_since_epoch()).count();
+        }
+        buffer->pts += data->producer_start_time - data->first_pts;
     }
 
     if (!gst_buffer_map(buffer, &info, GST_MAP_READ)){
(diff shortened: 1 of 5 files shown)


kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
```diff
@@ -470,6 +477,14 @@ static GstFlowReturn on_new_sample(GstElement *sink, CustomData *data) {
         if (CHECK_FRAME_FLAG_KEY_FRAME(kinesis_video_flags)) {
             data->key_frame_pts = buffer->pts;
         }
+    } else if (data->use_absolute_fragment_times) {
```
MushMal: I can't quite understand this logic. Could you please add come comments? Seems that we are trying to start absolute cluster times but from 0?

Also, LN 446 - this is an odd if statement.

if (!(dts_sign > 0))

should be

if (dts_sign <= 0)

zhiyua-git: The logic is setting first frame pts as first_pts and producer_start_time as current time when the first frame arrives. Then with each frame do a minus on first_pts to get a delta between it and first frame’s pts and add producer start time to it to form the absolute timestamp. Thought the naming would be enough so I didn’t add the comment, do you see any specific name make it hard to understand or some of condition or assignments make it hard? I can improve those, or if not any, I can add a general comment also.

MushMal: Perhaps I was not clear. The part you've described made sense. The question is why in absolute timestamp mode we need to keep the "presentation" start time and add it to the frame delta from the first frame? If we rely on the media clock then all should just go as is right?

zhiyua-git: There is no guarantee that gstreamer will return us with first pts as 0. Did some search and found someone reporting pts not starting from 0 before, not sure if it is a bug or acceptable behavior of gstreamer. https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/issues/594

MushMal: I guess, the main question to ask is what we really think of as absolute time on the producer device. We could simply rely on whatever the media clock states and timestamp that - after all, this is what's reported by the media pipeline. However, we need to be careful to ensure that all media pipelines are timestamped with the same clock (need to check). Then, we need to ensure the customers understand the behavior when the device is restarted and when a new streaming session happens. For example, if the actual reported clock is the wall clock then regardless of the streaming session we would have proper timeline - assuming the producer clock doesn't drift (some devices would have NTP running which would adjust the clock on boot up). Perhaps, we should simply take a start timestamp from the customer to timestamp against that or if the timestamp is not specified just use whatever the media pipeline reports?

zhiyua-git: This logic is relying on the media pipeline clock assuming it handles time and should be starting from 0 timestamp drift from producer start time(current time). I think it is same as what you mentioned latest.

kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
```diff
@@ -630,11 +645,11 @@ void kinesis_video_init(CustomData *data) {
 void kinesis_video_stream_init(CustomData *data) {
 
     STREAMING_TYPE streaming_type = DEFAULT_STREAMING_TYPE;
-    bool use_absolute_fragment_times = DEFAULT_ABSOLUTE_FRAGMENT_TIMES;
+    data->use_absolute_fragment_times = DEFAULT_ABSOLUTE_FRAGMENT_TIMES;
```
chehefen: we dont really need this use_absolute_fragment_times variable. Previously we are using relative mode for realtime and absolute mode for offline. Now they are both absolute.

kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
```diff
@@ -158,6 +160,11 @@ typedef struct _CustomData {
 
     uint32_t total_track_count;
 
+    bool use_absolute_fragment_times;
+
+    // Pts of first frame
+    uint64_t first_pts;
```
chehefen: should be per track in case of audio video. may happen that audio video use different clocks.

kinesis-video-gst-demo/kinesis_video_gstreamer_sample_multistream_app.cpp
```diff
@@ -176,6 +176,9 @@ typedef struct _CustomData {
     map<string, bool> stream_started;
     map<string, uint8_t*> frame_data_map;
     map<string, UINT32> frame_data_size_map;
+    // Pts of first frame
+    map<string, uint64_t> first_pts_map;
+    map<string, uint64_t> producer_start_time_map;
```
chehefen: This sample is meant for streaming multiple RTSP streams at the same time. In that case I think the start time should be the same for all streams.

kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
zhiyua-git: I think we still want it as offline will always go absolute while live can be modify to support relative mode by changing the DEFAULT_ABSOLUTE_FRAGMENT_TIMES to FALSE without affecting offline mode if relative time mode is needed for some reason.

kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
zhiyua-git: We were relying on the audio video getting sync clock from gstreamer before, if setting both  to start from time diff 0 then the audio video frame can have timestamp overlapping and failure in producer sdk.

kinesis-video-gst-demo/kinesis_video_gstreamer_sample_multistream_app.cpp
zhiyua-git: Shouldn't it be unique for each RTSP stream? The frames coming from different RTSP url might not start flowing at the same start time, the compensation will force all the RTSP stream starting from first one's start time if we only keep one producer_start_time.

chehefen: make sense

kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
chehefen: ok

chehefen:
approved

Issue #377, opened 2020-03-23
Auth.cpp in CPP producer can generate stale credentials making PIC to fail


MushMal:
Auth.cpp LN 34: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/kinesis-video-producer/src/Auth.cpp#L34

Due to the fact that PIC adds random jitter, the credentials stored here might become stale causing PIC client state machine Auth integration to fail. This will be primarily exposed as a putFrame error with status code: STATUS_CLIENT_AUTH_CALL_FAILED: x52000040.

PR #378, opened 2020-03-24
Fix the static credential provider caching logic with a potential of …

MushMal:
…returning stale credentials due to jitter

    Issue: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/377

Issue #377 

*Description of changes:*


kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
```diff
@@ -56,7 +56,7 @@ LOGGER_TAG("com.amazonaws.kinesis.video.gstreamer");
 #define APP_SINK_BASE_NAME "appsink"
 #define DEFAULT_BUFFER_SIZE (1 * 1024 * 1024)
 #define DEFAULT_STORAGE_SIZE (128 * 1024 * 1024)
-#define DEFAULT_CREDENTIAL_ROTATION_SECONDS 2400
+#define DEFAULT_CREDENTIAL_ROTATION_SECONDS 3600
```
disa6302: How does changing creds rotation from 40 minutes to 60 minutes help?

MushMal: This is not the real root cause but rather arbitrary value. Most of the session tokens issued have an hour life span so it's better to mimmic it.

PR #381, opened 2020-03-25
fix issue when sample upload a file, producer start time is 0.

chehefen:
*Issue #, if available:*

*Description of changes:*


kinesis-video-gst-demo/kinesis_video_gstreamer_audio_video_sample_app.cpp
```diff
@@ -481,9 +481,6 @@ static GstFlowReturn on_new_sample(GstElement *sink, CustomData *data) {
         if (data->first_pts == GST_CLOCK_TIME_NONE) {
             data->first_pts = buffer->pts;
         }
-        if (data->producer_start_time == GST_CLOCK_TIME_NONE) {
```
MushMal: So, in case it does start from 0 do we still do absolute times or relative? In case of absolute we will end up with 1970s timestamp. Should we compensate?

chehefen: We never start from 0. It is initialized to current time.

Issue #424, opened 2020-04-23
Parse parameters from json file in samples


disa6302:
Add more control to modify different parameters that control streams in samples.

disa6302:
Being tracked in producer C repo for now

PR #425, opened 2020-04-23
Add parameter provision in gstreamer samples

disa6302:
*Issue #, if available: #424 *

*Description of changes:*
- Added a json file with parameters that will be parsed by the apps 
- Added `parse` command line option to allow parameters from json file to be parsed
- Added a json file generator script to generate the json file


chehefen:
do you think it would make sense to moving the parsing functionality into c producer and give it ability to generate streamInfo based on the config file? Then this functionality can also be exposed through cpp layer.

disa6302:
@chehefen , I did initially think about having the parsing functionality in the producer C and exposing it through the cpp layer. But, the parser works on different args (Custom data in producer cpp and Sample custom data in producer C). What I can do is create a new config_data struct and and add it as member of custom data structure in both to make this a generic functionality. Let me go ahead and add this functionality in the Util.c in producer C

chehefen:
The new variables you added in custom data are mirrors for fields in stream_definition. You can add a new streamInfoProvider constructor in c producer that parses the config file and set up its member. The add another new api in StreamDefinition too that just calls the c producer api. Then in the sample if use_parsed, call the new api to setup streamDefinition.

MushMal:
> The new variables you added in custom data are mirrors for fields in stream_definition. You can add a new streamInfoProvider constructor in c producer that parses the config file and set up its member. The add another new api in StreamDefinition too that just calls the c producer api. Then in the sample if use_parsed, call the new api to setup streamDefinition.

What do others thing? I like this.

disa6302:
Will add this to producer C and reference here. So closing this PR for now.

PR #439, opened 2020-05-04
Update openssl version to 1.1.1g

disa6302:
*Issue #, if available:*

*Description of changes:*


disa6302:
I have a branch for producer C openssl update. 

For WebRTC, I do not know. Are there any other dependencies that depend on specific openssl version?

MushMal:
> I have a branch for producer C openssl update.
> 
> For WebRTC, I do not know. Are there any other dependencies that depend on specific openssl version?

Our codebase doesn't rely on a specific version. However, you are correct that other dependencies might. For example, SCTP. I would suggest to simply run it and see whether it produces any issues - normally a minor rev of the version shouldn't impact it.

disa6302:
I will test on WebRTC and then make a change. Merging this now.

PR #442, opened 2020-05-06
Add a file uploader example

lherman-cs:
*Issue #, if available:*

*Description of changes:*


CMakeLists.txt
```diff
@@ -200,6 +200,9 @@ if(BUILD_GSTREAMER_PLUGIN)
 
   add_executable(kinesis_video_gstreamer_audio_video_sample_app samples/kinesis_video_gstreamer_audio_video_sample_app.cpp)
   target_link_libraries(kinesis_video_gstreamer_audio_video_sample_app ${GST_APP_LIBRARIES} KinesisVideoProducer)
+
```
MushMal, line 203: Let's try to move this directly to the samples.

CMakeLists.txt
```diff
@@ -200,6 +200,9 @@ if(BUILD_GSTREAMER_PLUGIN)
...
+  add_executable(kinesis_file_gstreamer_sample_app samples/kinesis_file_gstreamer_sample_app.cpp)
```
MushMal, line 204: Need to think about the naming strategy. Prefixing everything with kinesis_video_streams_ is way to long.

Let's use kvs_

in this case it will be kvs_gstreamer_file_upload_sample

Obviously, you are free to chose a better naming.

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+extern "C" {
+#endif
+
+int gstreamer_init(int, char **);
+
+#ifdef __cplusplus
+}
+#endif
+
+LOGGER_TAG("com.amazonaws.kinesis.video.gstreamer");
```
MushMal, line 26: Consider changing the tag to something shorter and specific to the app. Can simply be the app name

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+
+#define DEFAULT_RETENTION_PERIOD_HOURS 2
```
MushMal, line 28: Once we are moving all of these to the samples repository, I would highly recommend having a Common directory which has a set of sources and definitions which will be shared among the samples. In this case, all of these definitions will go to Common.h or something

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    char *stream_name;
+
+    mutex audio_video_sync_mtx;
+    condition_variable audio_video_sync_cv;
+
+    // file to upload
+    FileInfo file;
+
+    // indicate whether a video key frame has been received or not.
+    volatile bool first_video_frame;
```
MushMal, line 95: Might want atomic_bool here too

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    // Unit: ns
+    uint64_t key_frame_pts;
+
+    // Assuming frame timestamp are relative. Add producer_start_time to each frame's
+    // timestamp to convert them to absolute timestamp. This way fragments dont overlap after token rotation when doing
+    // file uploading.
+    uint64_t producer_start_time;
+
+    unique_ptr<Credentials> credential;
+} CustomData;
```
MushMal, line 119: BTW, why not moving this into the namespace

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    static STATUS
+    droppedFrameReportHandler(UINT64 custom_data, STREAM_HANDLE stream_handle,
+                              UINT64 dropped_frame_timecode);
+
+    static STATUS
+    fragmentAckReceivedHandler( UINT64 custom_data, STREAM_HANDLE stream_handle,
+                                UPLOAD_HANDLE upload_handle, PFragmentAck pFragmentAck);
+};
+
+class SampleCredentialProvider : public StaticCredentialProvider {
```
MushMal, line 180: Not sure if this really adds any value. Perhaps we should use StaticCredentialProvider directly. I think this was an attempt to demonstrate how a custom credential provider might work. However, once these are outsourced into the Common/ directory we can think of adding it back or something. Let me know what you think

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    frame.size = static_cast<UINT32>(len);
+    frame.frameData = reinterpret_cast<PBYTE>(data);
+    frame.trackId = track_id;
+    return kinesis_video_stream->putFrame(frame);
+}
+
+bool all_stream_started(CustomData *data) {
+    bool started = true;
+    if (data->stream_started.size() < DEFAULT_TOTAL_TRACK_COUNT) {
+        started = false;
```
MushMal, line 302: Very minor. feel free to completely ignore: This is a simple function so in my opinion it's better to just return false from here.

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    } else {
+        for (map<int, bool>::iterator it = data->stream_started.begin(); it != data->stream_started.end(); ++it) {
+            if (!it->second) {
+                started = false;
```
MushMal, line 306: same here

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+                break;
+            }
+        }
+    }
+
+    return started;
```
MushMal, line 312: in that case, just return true here

samples/kinesis_file_gstreamer_sample_app.cpp
```diff
@@ -0,0 +1,797 @@
...
+    if (!gst_element_link_many(audio_queue, aacparse, audio_filter, audio_appsink,
+                               NULL)) {
+        g_printerr("Audio elements could not be linked.\n");
+        gst_object_unref(pipeline);
+        return 1;
+    }
+
+    return 0;
+}
+
```
MushMal, line 677: extra \n

CMakeLists.txt
lherman-cs: Do you mean the samples repo, https://github.com/aws-samples/amazon-kinesis-video-streams-demos/tree/samples-folder? 

If yes, I'm also working on it rn. I'm planning to merge this PR, then later transfer it to the demo repo.

CMakeLists.txt
lherman-cs: yea! I definitely feel the same way.

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: Makes sense!

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: Yep, that makes sense, similar to what we have in webrtc sdk

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: No, it's alright since it's already protected by the mutex, https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/442/files#diff-2af4741aceb98d75d1fb60eac19ef4f3R316

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: No, it's alright since it's already protected by the mutex, https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/442/files#diff-2af4741aceb98d75d1fb60eac19ef4f3R316

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: Yea, I think this a good idea, I think users don't need to be burdened to read this part since this is not the main focus.

samples/kinesis_file_gstreamer_sample_app.cpp
lherman-cs: IMO, I think it's not appropriate since this is an application scope, not the library scope.

But, I honestly don't have a hard feeling either way, please let me know what you think.

lherman-cs:
As @chehefen suggested offline, we can replace the pipeline creation with gstreamer launch instead to make it simpler.

lherman-cs:
Updated with kvssink gstreamer plugin. Following are environment variables you can configure for the sample

Screen Shot 2020-05-08 at 6 36 05 PM

disa6302:
I am wondering if we need the env option for all the parameters. I have an open PR in producer C SDK to parse different stream and device properties from a json file. You or I could add those changes into this sample once that goes in. Let me know what you think.

lherman-cs:
> I am wondering if we need the env option for all the parameters. I have an open PR in producer C SDK to parse different stream and device properties from a json file. You or I could add those changes into this sample once that goes in. Let me know what you think.

I think it's fine to keep them separated. My goal with this sample is that to have everything self contained, minimize the number of dependencies (`gst/gst.h` is the only header dependency + our kvssink gstreamer plugin), and should be super simple.

I think the library users (or application level) should decide what dependencies or conventions that they want to use, for example `json` is probably not appropriate for their application and probably they don't know how to use jsmn.

disa6302:
> > I am wondering if we need the env option for all the parameters. I have an open PR in producer C SDK to parse different stream and device properties from a json file. You or I could add those changes into this sample once that goes in. Let me know what you think.
> 
> I think it's fine to keep them separated. My goal with this sample is that to have everything self contained, minimize the number of dependencies (`gst/gst.h` is the only header dependency + our kvssink gstreamer plugin), and should be super simple.
> 
> I think the library users (or application level) should decide what dependencies or conventions that they want to use, for example `json` is probably not appropriate for their application and probably they don't know how to use jsmn.

The application never needs to know how to use jsmn because it all happens under the hood. But, I am fine with having them separated :) . Just threw in a suggestion.

lherman-cs:
Oh, right, I just realized that the parsing is done in the library source, I thought it was in the sample. But, that also means that it should live inside of `kvssink` gstreamer plugin, which I don't know if that will feel appropriate 🙂.

PR #450, opened 2020-05-13
Reduce available properties for file uploader sample

lherman-cs:
* Remove unnecessary properties
* Remove redudant properties
* Convert stream name and file start time to be arguments

### DIFF
--- a/samples/kvs_gstreamer_file_uploader_sample.cpp
+++ b/samples/kvs_gstreamer_file_uploader_sample.cpp
@@ -17,6 +17,7 @@ int gstreamer_init(int, char **);
 #endif
 
 #define DEFAULT_RETRY_COUNT       3
+#define DEFAULT_STREAMING_TYPE    "offline"
 
 #define PROPERTY_PREFIX           "KVS_"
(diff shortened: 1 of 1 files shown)


samples/kvs_gstreamer_file_uploader_sample.cpp
```diff
@@ -240,20 +238,32 @@ int main(int argc, char* argv[]) {
...
     }
-    data.file_path = argv[1];
-    data.kvssink_str = build_kvssink_str();
+    string stream_name = argv[1];
+    data.file_path = argv[2];
+    if (access(data.file_path.c_str(), F_OK) == -1) {
+        LOG_ERROR("File not found. Path: %s", data.file_path.c_str());
+        return 1;
+    }
+    unsigned long file_start_time = strtoul(argv[3], &ptr, 10);
```
chehefen, line 253: if file_start_time is 0 then use current time?

lherman-cs: Yep

PR #495, opened 2020-06-22
Remove submodule. Add script to pull down c producer and build with add_subdirectory

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CMake/Utilities.cmake
+++ b/CMake/Utilities.cmake
@@ -1,3 +1,34 @@
+# only fetch target repo for add_subdirectory later
+function(fetch_repo lib_name)
+  set(supported_libs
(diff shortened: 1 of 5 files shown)


CMake/Utilities.cmake
```diff
@@ -1,3 +1,35 @@
+# only fetch target repo for add_subdirectory later
+function(fetch_repo lib_name)
+  set(supported_libs
+      kvspic
```
disa6302, line 4: There is no CMakeLists equivalent file for this. We can remove it.

chehefen: yup. done.

Issue #494, opened 2020-06-22
[BUG] CMake error with "target_compile_definitions" and environment variables


Subburaj47:
**Describe the bug**
I believe the minimum_required_version in the available CMakes are 2.8. I face a problem with the cmake while cmake runs in 2.8 versions. 

1. The cmake function "target_compile_definitions" seems to be an unknown command.
2. "CMAKE_LANGUAGES_COMPILER_ENV_VAR" and "CMAKE_LANGUAGES_COMPILER" modules are not found.

**To Reproduce**
Have a 2.8 version of cmake available in the build system. Checkout the 3.0.0 SDK and trigger the cmake. 

**Expected behavior**
CMake would have got executed peacefully and subsequently the build.

**SDK version number**
SDK version used is 3.0.0

**Platform (please complete the following information):**
running in linux.

**Additional context**
Any option other than the cmake version upgrades would help
[amazon-kinesis-video-streams-producer-sdk-cpp.log](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/files/4813699/amazon-kinesis-video-streams-producer-sdk-cpp.log)
.

Sean-Der:
@Subburaj47 Thanks for the report!

What Linux distro + version are you using? Will reproduce this today, thanks!

Subburaj47:
@Sean-Der : Running on a thirdparty system. Here are the details.
/mnt/ramdisk/root # cat /proc/version
Linux version 3.10.104+ (charlie@isbu-virtual-machine) (gcc version 5.2.1 20151005 (Linaro GCC 5.2-2015.11-2) ) #1 PREEMPT Mon May 18 16:01:21 CST 2020

MushMal:
Seems that CMake 2.8.11 contains target_compile_definitions

https://cmake.org/cmake/help/v2.8.11/cmake.html#command:target_compile_definitions

and this is the earliest version that started to have the directive.

We should update the make file to reflect this

MushMal:
This is fixed

PR #498, opened 2020-06-29
Update cmake version to be more specific

lherman-cs:
Issue #494 

The PR will add a specific supported CMake version to the CI so that if we accidentally use a feature that's not available in the supported version, the CI will fail.  

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ matrix:
         # At least one of jobs should use a specific version of cmake so that we know the lower bound version
         - cmake --version
         - sudo rm `which cmake`
-        - sh -c "wget https://cmake.org/files/v3.12/cmake-3.12.4.tar.gz && tar -xf cmake-3.* && cd cmake-3.12.4 && ./bootstrap && make -j4 && sudo make install"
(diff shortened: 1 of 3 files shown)


lherman-cs:
@MushMal I'm not really sure. But, to be safe, I'll add this to C Producer and WebRTC so that we can always know if we accidentally use a feature that's not available in the current cmake version.

.travis.yml
```diff
@@ -54,6 +54,12 @@ matrix:
       os: linux
       compiler: gcc
       before_script:
+        # At least one of jobs should use a specific version of cmake so that we know the lower bound version
+        - cmake --version
+        - sudo rm `which cmake`
+        - sh -c "wget https://cmake.org/files/v3.12/cmake-3.12.4.tar.gz && tar -xf cmake-3.* && cd cmake-3.12.4 && ./bootstrap && make -j4 && sudo make install"
```
lherman-cs, line 60: Unfortunately, 3.12.4 is the lowest common denominator. Although, if we look at our dependencies, the highest requirement is 3.0. But, for whatever reason, it couldn't find OpenSSL properly when I used 3.0.

Issue #538, opened 2020-08-11
[BUG] Use of `using namespace std;` and `using namespace std::chrono;` should be forbidden and breaks this code's ability to be used in many cases


emersonknapp:
**Describe the bug**
Headers in this library use `using namespace std;` and `using namespace std::chrono;` (see example https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/src/StreamDefinition.h#L18) 

This means that, for example, when I link against a library that uses boost's `time_point`, my code cannot compile because of ambiguous definitions. It is C++ best practice to never `using namespace` in a header because it can inadvertently change the meaning of downstream code, and its effects cannot be undone by the consuming code.

**To Reproduce**

**Expected behavior**

**SDK version number**
Include the SDK version you are running. If it is a specific commit on master branch, include that

**Open source building**
If it is a build issue, include 3rd party library version and steps to how you are building it

**Screenshots**

**Platform (please complete the following information):**
 - OS: [e.g. Linux, Windows]
 - Processor: [e.g. ARMv5]
 - Version [e.g. 18.04]

**Additional context**

MushMal:
+1 on this. @emersonknapp do you want to to take a stab at it? We would love to get public contributions to the codebase

chehefen:
closing this issue as it has been fixed in #544

PR #541, opened 2020-08-12
Remove 'using namespace std' from all headers

emersonknapp:
Closes https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/538

Remove `using namespace std` from all headers in the project, so as to not irreversibly modify the global namespace for users of this library.

### DIFF
--- a/samples/kvs_gstreamer_audio_video_sample.cpp
+++ b/samples/kvs_gstreamer_audio_video_sample.cpp
@@ -15,6 +15,7 @@
 #include <queue>
 
 using namespace std;
+using namespace std::chrono;
 using namespace com::amazonaws::kinesis::video;
 using namespace log4cplus;
 
--- a/samples/kvs_gstreamer_multistream_sample.cpp
+++ b/samples/kvs_gstreamer_multistream_sample.cpp
@@ -9,6 +9,7 @@
 #include <map>
 
 using namespace std;
+using namespace std::chrono;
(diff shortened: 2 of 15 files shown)


emersonknapp:
@mushmal requesting review

emersonknapp:
Ah - I didn't have the gstreamer plugin enabled in my local build

MushMal:
> Ah - I didn't have the gstreamer plugin enabled in my local build

Thanks for taking care of it. The changes look good. Let's wait for the clean build. I am pretty sure the tests will still fail due to the need to have credentials to run against the backend. We will address that by re-submitting the PR

MushMal:
Few more compile errors.

emersonknapp:
Ok looks like tests also don't build by default - noting that I am now using configuration

```
cmake .. -DBUILD_GSTREAMER_PLUGIN=ON -DBUILD_JNI=OFF -DBUILD_DEPENDENCIES=ON -DBUILD_TEST=ON
```

emersonknapp:
@MushMal from what I can tell it is now test failures due to credentials, is that correct?

MushMal:
> @MushMal from what I can tell it is now test failures due to credentials, is that correct?

Oh yes, these are auth failures. Let us take care of it and push the commit

emersonknapp:
Resolved by https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/544

PR #654, opened 2021-02-08
fix producer cpp set correct default for intermittent producer scenar…

hassanctech:
…io and docs

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/intermittent_producer.md
+++ b/docs/intermittent_producer.md
@@ -1,5 +1,14 @@
-### Intermittent Producer Handling
-Since v3.1.0 of the Producer C++ SDK by default we have automatic handling for intermittent streaming.  This comes at a cost of using 1 extra thread to manage the timer used by the SDK to automatically signal to the KVS back-end that a fragment can be closed out.  If on a prior release of the SDK you were manually setting fields like `stream_info_.streamCaps.frameOrderingMode` you no longer should do that.  An example of this scenario is if you stream for an amount of time, then you stop producing frames (for example the case of a security camera which is only uploading video/audio during a motion event) and then after an unknown amount of time (for example the next motion event) you start streaming again.  This is all automatically handled by the SDK now.
+## What is intermittent streaming?
+Intermittent streaming is when you want to send and stop video at will.  For example consider a doorbell camera which only wants to record motion events.  The application might want to start producing frames as soon as it detects motion and then stop at some point when there is no longer motion and then not stream again until it detects motion. 
+
+## Intermittent Producer Handling
+### Before v3.1.0
+The KVS back-end times out after 30s of not receiving any frames.  If the SDK received frames after 30+ seconds of no frames then it would error.  We introduced a complex way for client applications to be able to make an API call which would signal to the backend that we're done recording and to close out the session.  However this does not work in cases where the stream application doesn't know that it's "done" streaming and it also doesn't know in advance whether or not it will need to stream again within the next 30s. 
+
+
+## Since v3.1.0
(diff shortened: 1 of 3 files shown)


src/StreamDefinition.cpp
```diff
@@ -108,7 +108,7 @@ void StreamDefinition::addTrack(const uint64_t track_id,
                                 MKV_TRACK_INFO_TYPE track_type,
                                 const uint8_t* codecPrivateData,
                                 uint32_t codecPrivateDataSize) {
-    stream_info_.streamCaps.frameOrderingMode = FRAME_ORDERING_MODE_MULTI_TRACK_AV_COMPARE_PTS_ONE_MS_COMPENSATE;
+    stream_info_.streamCaps.frameOrderingMode = FRAME_ORDERING_MODE_MULTI_TRACK_AV_COMPARE_PTS_ONE_MS_COMPENSATE_EOFR;
```
sharadrj, line 111: These modes are inscrutable for customers :(
Not needed for this PR, but eventually we should remove frameOrderingMode altogether as a customer controlled setting.

docs/intermittent_producer.md
```diff
@@ -0,0 +1,12 @@
+### Intermittent Producer Handling
+Since v3.1.0 of the Producer C++ SDK by default we have automatic handling for intermittent streaming.  This comes at a cost of using 1 extra thread to manage the timer used by the SDK to automatically signal to the KVS back-end that a fragment can be closed out.  If on a prior release of the SDK you were manually setting fields like `stream_info_.streamCaps.frameOrderingMode` you no longer should do that.  An example of this scenario is if you stream for an amount of time, then you stop producing frames (for example the case of a security camera which is only uploading video/audio during a motion event) and then after an unknown amount of time (for example the next motion event) you start streaming again.  This is all automatically handled by the SDK now.
```
sharadrj, line 2: Recommend starting the paragraph by defining what we mean by intermittent streaming, then telling how clients and SDK work in this scenario today, and lastly telling what exactly is "automatically" handled in v3.1. Current flow is roughly backwards, which makes it difficult to create a mental picture of what is happening

src/StreamDefinition.cpp
MushMal: Here we actually hard-code the value rather let the application supply it

MushMal (approved):

src/StreamDefinition.cpp
hassanctech: Right in this PR I'm just updating the hardcoded value but I think @sharadrj is talking about just a few lines down we do have a public API method:      

void setFrameOrderMode(FRAME_ORDER_MODE mode);

We should remove this in the future perhaps in favor of a more user understandable enum which we can document an then internally we can use that value to pick the correct frame order mode.

docs/intermittent_producer.md
hassanctech: Makes sense, see new commits.

docs/intermittent_producer.md
```diff
@@ -0,0 +1,21 @@
+## What is intermittent streaming?
+Intermittent streaming is when you want to send and stop video at will.  For example consider a doorbell camera which only wants to record motion events.  The application might want to start producing frames as soon as it detects motion and then stop at some point when there is no longer motion and then not stream again until it detects motion.
```
sharadrj, line 2: >Intermittent streaming is when you want to send and stop video at will

Should clarify that sending and stopping video at will *without tearing and setting up SDK pipeline again*

disa6302 (approved):

PR #658, opened 2021-02-11
provide option for non-standard GStreamer src/mux plugins which have …

hassanctech:
Introduces new `kvssink` property `disable-buffer-clipping`.  This defaults to false and most users will not need to set it.  However some src/mux elements produce non-standard segment start times.  If you are aware that your src/mux elements do this then you can try to set this value to true.

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -89,6 +89,7 @@ GST_DEBUG_CATEGORY_STATIC (gst_kvs_sink_debug);
 #define DEFAULT_FRAGMENT_ACKS TRUE
 #define DEFAULT_RESTART_ON_ERROR TRUE
 #define DEFAULT_RECALCULATE_METRICS TRUE
+#define DEFAULT_DISABLE_BUFFER_CLIPPING FALSE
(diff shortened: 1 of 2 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -380,7 +382,7 @@ void create_kinesis_video_stream(GstKvsSink *kvssink) {
     if (data->media_type == AUDIO_VIDEO) {
         stream_definition->addTrack(KVS_SINK_DEFAULT_AUDIO_TRACKID, DEFAULT_AUDIO_TRACK_NAME, kvssink->audio_codec_id, MKV_TRACK_INFO_TYPE_AUDIO);
         // Need to reorder frames to avoid fragment overlap error.
-        stream_definition->setFrameOrderMode(FRAME_ORDERING_MODE_MULTI_TRACK_AV_COMPARE_PTS_ONE_MS_COMPENSATE);
+        stream_definition->setFrameOrderMode(FRAME_ORDERING_MODE_MULTI_TRACK_AV_COMPARE_PTS_ONE_MS_COMPENSATE_EOFR);
```
MushMal, line 385: Niiice

MushMal (approved):

PR #660, opened 2021-02-12
Move check out of switch

hassanctech:
With the previous commit the behavior was to *always* disable clipping unless the property was explicitly set, so default was not honored.  Moved the clip function set up to the producer init which happens *after* ALL of the properties have been read in.  Tested both with and without defaults and explicitly set parameter, works as expected

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -89,6 +89,7 @@ GST_DEBUG_CATEGORY_STATIC (gst_kvs_sink_debug);
 #define DEFAULT_FRAGMENT_ACKS TRUE
 #define DEFAULT_RESTART_ON_ERROR TRUE
 #define DEFAULT_RECALCULATE_METRICS TRUE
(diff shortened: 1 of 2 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -249,6 +249,12 @@ void kinesis_video_producer_init(GstKvsSink *kvssink)
     string region_str;
     bool credential_is_static = true;
 
+    // This needs to happen after we've read in ALL of the properties
+    if (kvssink->disable_buffer_clipping == FALSE) {
```
MushMal, line 253: Don’t compare Boolean against literal. Do if (!blah) {}

MushMal (approved):
Minor comment.

src/gstreamer/gstkvssink.cpp
hassanctech: I really didn't see the reason to do this but you're right, for other people's benefit:
https://developer.gnome.org/glib/unstable/glib-Basic-Types.html#gboolean

PR #719, opened 2021-07-06
Check system for cproducer before doing git clone

spotaws:
*Issue #, if available:*

*Description of changes:*
Currently, the CMakeLists.txt assumes that you do not have the kvscproducer library and does a git checkout of https://github.com/awslabs/amazon-kinesis-video-streams-producer-c in order to meet the required dependency. This makes producer-sdk-cpp much harder to package (e.g. RPM/DEB) as most distributions do not permit network access during build and this forced git call will fail.

This patch checks for the kvscproducer library dependency that producer-sdk-cpp cares about (libcproducer) and uses it if found. A more specific check could be done here to ensure that the found library has the necessary functionality, but to be honest, it would be easier to use a proper versioning scheme for kvscproducer and just check for supported version.

If the kvscproducer library is not found, it falls back to doing the git checkout logic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,17 +84,24 @@ set(BUILD_COMMON_LWS
 set(BUILD_COMMON_CURL
     TRUE
     CACHE BOOL "Build ProducerC with CURL Support" FORCE)
-set(DEPENDENCY_DOWNLOAD_PATH ${CMAKE_CURRENT_SOURCE_DIR}/dependency)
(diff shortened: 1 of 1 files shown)


hassanctech:
I've migrated this PR here:  https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1115

@spotaws Your commit is there, which will run through our CI.  Apologies for the (long) delay.

PR #721, opened 2021-07-06
install some files

spotaws:
*Issue #, if available:*

*Description of changes:*
So, um, cmake does not actually install anything right now. I understand that this code is in active development, but still... This change installs the binaries, the gstreamer plugin, and the JNI header files into what I think are correct targets. There are other header files in the tree that probably need to be installed, but I wasn't sure where they would go.


niyatim23 (commented):
I just noticed that https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1087 covers a part of this. Not sure if we want to install JNI and our samples as well though

niyatim23 (commented):
Closing in favor of https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1118. Thanks for your contribution!

PR #723, opened 2021-07-06
make versioned shared libraries

spotaws:
*Issue #, if available:*

*Description of changes:*
This change creates versioned shared libraries for the two true shared libraries. The gstreamer plugin is a plugin, not a shared library, so it is left unversioned.


disa6302:
Thank you! We have pulled in your commit in #1116 . Closing this.

PR #746, opened 2021-08-31
added missing finalize property

waikup83:
memory leak detected

added property to be freed when the object has finalized


MushMal (approved):

hassanctech (approved):
Thanks for this!

PR #792, opened 2021-11-15
Warn when frames are dropped

daveisfera:
*Issue #, if available:*
#768 
 
*Description of changes:*
Log dropped frames as a `WARN` instead of `DEBUG`


daveisfera:
Spewing the terminal? Shouldn't this be an infrequent event? And if frames are dropping, then shouldn't the user know about it instead of it being silently ignored?

disa6302:
It could spew the terminal. I am concerned about making it DEBUG.  It is supposed to be infrequent, but say it runs for a long time without user noticing, then it could fill up the logs.

PR #793, opened 2021-11-15
Update version of openssl that's built

daveisfera:
*Issue #, if available:*
N/A

*Description of changes:*
Update to the latest version of openssl for good measure

### DIFF
--- a/CMake/Dependencies/libopenssl-CMakeLists.txt
+++ b/CMake/Dependencies/libopenssl-CMakeLists.txt
@@ -21,7 +21,7 @@ endif()
 include(ExternalProject)
 ExternalProject_Add(project_libopenssl
     GIT_REPOSITORY    https://github.com/openssl/openssl.git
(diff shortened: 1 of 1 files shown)


daveisfera:
There are 7 CVEs and a few other security fixes, so just seems like a good idea and the right thing to do

hassanctech:
We can consider this however we've found even around our selected version of openssl there were adjacent ones that for example didn't build on ARM platform, we will look into this and see if there are any issues.  We must support mips as well so will need to test that one also.  If you have tested our SDK with this version on a mips/arm platform and had no issues that datapoint would be helpful feel free to share!

daveisfera:
We're currently only building and running on x86

disa6302:
Updated to 1.1.1t here: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/950

Issue #878, opened 2022-07-08
[QUESTION] What is the alternative of omxh264enc for Raspberry PI os (bulleye)?


LLejoly:
How are we supposed to use the SDK to stream real time video with the Raspberry PI OS bullseye which is the most recent version of the OS?

The omxh264enc is not supported anymore. Thus what is the alternative

** Logging **
omxh264enc not supported anymore with PI OS Bullseye

** Any design considerations/constraints **
- Use of PI OS Bullseye
- The alternative proposed with x264enc creates an allocation error.

Explain in detail how you would like to integrate our SDK into your solution 

The goal is to find a way to integrate the kvssink on raspberry pi OS (Bullseye)

LLejoly:
Solution found:

With bullseye you need to use libcamerasrc to retrieve camera source and you need to use a supported h264enc.

```
gst-launch-1.0 libcamerasrc ! video/x-raw,width=640,height=480,framerate=30/1,format=I420 ! videoconvert ! v4l2h264enc extra-controls="controls,repeat_sequence_header=1" ! video/x-h264,level='(string)4' ! h264parse ! video/x-h264,stream-format=avc, alignment=au,width=640,height=480,framerate=30/1 ! kvssink stream-name="yourStreamName"  aws-key"yourKey" secret-key="yourSecretKey" aws-region="yourRegion"
```

PR #879, opened 2022-07-08
Support for Bullseye version

LLejoly:
*Issue #878 

*Description of changes:*

Provides a solution for the Raspberry PI OS (Bullseye) to stream data to kvs. The tutorial provided is outdated for Bullseye since encoders and source acquisition are not supported anymore.


zEuS0390 (approved):
This should be included in the documentation for the latest versions of Raspberry Pi OS. It's hard to find the solution when it is not included in the official.

Issue #909, opened 2022-09-23
[BUG] Static build fails


gregbreen:
**Describe the bug**
Building with BUILD_STATIC does not work. Several problems:

1. **libgstkvssink** is wrongly attempted to be linked as a shared object.
2. **log4cplus** is wrongly built as a shared object, causing samples to not link.
3. Use of `-static` to force a full static link causes a link failure because we are not linking against a full static build of gstreamer (**gstreamer-full-1.0**). Furthermore, a full static link is not advised for GNU C libraries and causes problems with Name Service Switch (NSS)
4. Link fails because cproducer is linked in the wrong order. Requires this PR: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/328

**Open source building**

```
mkdir build && cd build
cmake .. -DBUILD_GSTREAMER_PLUGIN=ON -DBUILD_STATIC=ON
make
```

**Platform (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04

**Additional context**

**libgstkvssink** is wrongly attempted to be linked as a shared object.

```
[ 94%] Linking CXX shared module libgstkvssink.so
/usr/bin/ld: /home/runner/work/amazon-kinesis-video-streams-producer-sdk-cpp/amazon-kinesis-video-streams-producer-sdk-cpp/open-source/local/lib/libcurl.a(setopt.c.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/gstkvssink.dir/build.make:169: libgstkvssink.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:171: CMakeFiles/gstkvssink.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
Error: Process completed with exit code 2.
```

**log4cplus** is wrongly built as a shared object, causing samples to not link.

```
[ 95%] Linking CXX executable kvs_gstreamer_sample
/usr/bin/ld: cannot find -lgstapp-1.0
/usr/bin/ld: cannot find -lgstbase-1.0
/usr/bin/ld: cannot find -lgstreamer-1.0
[...]

PR #910, opened 2022-09-23
Fix static build

gregbreen:
*Issue #, if available:*

#909

*Description of changes:*

Several changes:

1. Add static build to CI
2. **gstkvssink** should not be built as a shared object when doing a static build
3. **log4cplus** should not be built as a shared object when doing a static build
4. Don't force full static build; we can't since we don't link **gststreamer-full-1.0** and it's not advised for GCC anyway because of NSS.
5. **cproducer** links in the wrong order; need this PR: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/328


hassanctech (approved):
lgtm

PR #917, opened 2022-10-17
create new sample that uses kvssink

benwkim:
*Description of changes:*
1. Create new sample based off kvs_gstreamer_sample.cpp that uses kvssink instead of appsink
2. Add new argument (-runtime XXX(seconds)) for RTSP and livestreaming that terminates session after XXX seconds
3. Add SIGINT handling

*Additional changes:*
1. CI/CD checks were failing due to timeout so increased the timeout limit to 60 minutes.


samples/kvssink_gstreamer_sample.cpp
```diff
@@ -0,0 +1,845 @@
...
+       g_printerr("Invalid input arguments\n");
+       return 1;
+    }
+
+    LOG_DEBUG("Streaming with live source and width: " << width << ", height: " << height << ", fps: " << framerate
+                                                       << ", bitrateInKBPS" << bitrateInKBPS);
+
+    GstElement *source_filter, *filter, *kvssink, *h264parse, *encoder, *source, *video_convert;
+
+    /* create the elemnents */
```
unicornss, line 344: let us  remove the launch sample cmd as we are not using file sink here

jdelapla (changes requested):

samples/kvssink_gstreamer_sample.cpp
benwkim: Done

jdelapla (approved):

PR #957, opened 2023-02-20
Signals in develop

disa6302:
*Issue #, if available:*

- Add signal to listen in on persisted acks
- Add signal to listen in on stream errors
- Address static analysis results
- Allow user-agents property
- Allow setting stream timeout as a property
- Allow using original PTS for gstreamer buffers
- Provide flag to disable creating stream if describe call fails.

*Description of changes:*


src/gstreamer/KvsSinkStreamCallbackProvider.cpp
```diff
@@ -39,12 +53,21 @@ KvsSinkStreamCallbackProvider::droppedFrameReportHandler(UINT64 custom_data,
...
+    return STATUS_SUCCESS; // continue streaming
+}
+
 STATUS
 KvsSinkStreamCallbackProvider::streamLatencyPressureHandler(UINT64 custom_data,
                                                             STREAM_HANDLE stream_handle,
                                                             UINT64 current_buffer_duration) {
     UNUSED_PARAM(custom_data);
-    LOG_DEBUG("Reported streamLatencyPressure callback for stream handle " << stream_handle << ". Current buffer duration in 100ns: " << current_buffer_duration);
+    LOG_INFO("Reported streamLatencyPressure callback for stream handle " << stream_handle << ". Current buffer duration in 100ns: " << current_buffer_duration);
```
MushMal, line 70: LOG_WARN?

src/gstreamer/KvsSinkStreamCallbackProvider.cpp
```diff
@@ -22,9 +33,12 @@ KvsSinkStreamCallbackProvider::streamErrorReportHandler(UINT64 custom_data,
     LOG_ERROR("Reported stream error. Errored timecode: " << errored_timecode << " Status: 0x" << std::hex << status_code);
     auto customDataObj = reinterpret_cast<KvsSinkCustomData*>(custom_data);
 
-    // ignore if the sdk can recover from the error
-    if (!IS_RECOVERABLE_ERROR(status_code)) {
-        customDataObj->stream_status = status_code;
+    if(customDataObj != NULL) {
+        // ignore if the sdk can recover from the error
+        if (!IS_RECOVERABLE_ERROR(status_code)) {
```
MushMal, line 38: Can combine these two (though most compiler optimizers would do it for you)

src/KinesisVideoProducer.h
```diff
@@ -41,6 +41,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video {
  */
 #define CLIENT_STREAM_CLOSED_CALLBACK_AWAIT_TIME_MILLIS (10 + TIMEOUT_AFTER_STREAM_STOPPED + TIMEOUT_WAIT_FOR_CURL_BUFFER)
 
+#define CONTROL_PLANE_URI_ENV_VAR ((PCHAR) "CONTROL_PLANE_URI")
```
MushMal, line 44: Does this belong in the core or in the samples? Would be nice to have the same behavior as in the C Producer - whether elements like these are defined in the main include or in the samples.

src/KinesisVideoProducer.cpp
```diff
@@ -194,7 +195,9 @@ KinesisVideoProducer::~KinesisVideoProducer() {
 
 KinesisVideoProducerMetrics KinesisVideoProducer::getMetrics() const {
     STATUS status = ::getKinesisVideoMetrics(client_handle_, (PClientMetrics) client_metrics_.getRawMetrics());
-    LOG_AND_THROW_IF(STATUS_FAILED(status), "Failed to get producer metrics with: " << status);
+    if(STATUS_FAILED(status)) {
+        LOG_ERROR("Failed to get producer metrics with: " << status);
```
MushMal, line 199: This is a behavioral change and a dangerous one too. Understandably, this is just getting metrics but it might be worth having a way for the caller to distinguish a failed call.

src/KinesisVideoProducer.cpp
```diff
@@ -156,7 +156,8 @@ shared_ptr<KinesisVideoStream> KinesisVideoProducer::createStreamSync(unique_ptr
 
 void KinesisVideoProducer::freeStream(std::shared_ptr<KinesisVideoStream> kinesis_video_stream) {
     if (nullptr == kinesis_video_stream) {
-        LOG_AND_THROW("Kinesis Video stream can't be null");
+        LOG_ERROR("Kinesis Video stream can't be null");
+        return;
```
MushMal, line 160: Might be worth looking into how these APIs can communicate a failure to the caller. Similar to the next comment.

MushMal (commented):
Most comments are minor with the exception of the behavioral change in the public API.

CMakeLists.txt
```diff
@@ -1,13 +1,9 @@
...
-  set(CMAKE_OSX_DEPLOYMENT_TARGET ${CMAKE_SYSTEM_VERSION})
-endif()
-
 cmake_minimum_required(VERSION 3.6.3)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake;${CMAKE_MODULE_PATH}")
 include(Utilities)
 project(KinesisVideoProducerCpp)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
MushMal, line 6: Just curious, why requiring C++ 17?

disa6302: We are updating log4cplus, whose build fails with this update. We also have a couple of other projects that need this version.

src/KinesisVideoProducer.cpp
disa6302: I would prefer setting these to STATUS return type, for all of them to be in line with the producer and PIC SDKs. But will tackle that in another PR.

src/KinesisVideoProducer.h
disa6302: Using this only in kvssink as of now. In producer C, atleast in the samples, we are using callbacks that allow passing this. Ok either ways.

jdelapla (approved):

PR #983, opened 2023-03-07
Update path to gsstreamer pkg-config.exe

ghitabn:
make sure that CMakeLists.txt has the right path to gstreamer's pkg-config.exe:
- default:   PKG_CONFIG_EXECUTABLE "C:\\gstreamer\\1.0\\x86_64\\bin\\pkg-config.exe"
- updated: PKG_CONFIG_EXECUTABLE "C:\\gstreamer\\1.0\\msvc_x86_64\\bin\\pkg-config.exe")

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/build_windows.bat
+++ b/.github/build_windows.bat
@@ -1,6 +1,7 @@
 call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
 mkdir build
 cd build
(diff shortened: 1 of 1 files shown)


niyatim23:
Hi @ghitabn, the change you have made is not syntactically correct for the bat file. It is not necessary to include gstreamer this way because we install with choco for our CI builds.

PR #1015, opened 2023-05-12
Copy changes

richardgavel-ordinaryexperts:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -958,6 +958,7 @@ gst_kvs_sink_set_property(GObject *object, guint prop_id,
             }
             kvssink->metadata_tags = s ? gst_structure_copy(s) : NULL;
             break;
+        }
         default:
             G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
             break;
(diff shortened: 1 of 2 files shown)


hassanctech:
@richardgavel-ordinaryexperts Thank you for your contribution!

I have some concerns regarding the design.   The metadata tagging is a fragment level feature.  If it is exposed as a GStreamer element property now this is something that is set at the start and we make the call after calling start() on the stream so it happens only one time. 

The PIC level API though it is designed to be called at any time during active streaming. Since each fragment can have a unique tag, I think a better way to express this in GStreamer is as a custom event that the application puts on the GStreamer bus which `kvssink` then knows how to handle and it makes the call to the put event metadata API.

hassanctech:
You have a valid point there it would be useful to customers to support something like this from the command line.  I'm thinking this can live in conjunction with a custom event (that the client application places on the pipeline bus) which kvssink then reacts to and calls this API on the "current" fragment.  This would allow finer grained control to clients who wish to have such control.  Let me consider your implementation a bit more carefully and get back.

disa6302:
Closing this in favor of https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1122

PR #1019, opened 2023-05-30
fix(sample): move putEventMetadata after putFrame

sirknightj:
*Issue #, if available:*
When running the sample (`./kvs_gstreamer_sample test-stream ~/test.mp4`), we see the following error:
```
[ERROR] [30-05-2023 21:09:32:144.912 GMT] putKinesisVideoEventMetadata(): operation returned status code: 0x5200008c
[ERROR] [30-05-2023 21:09:32:144.981 GMT] Failed to insert event metadata with: 1375731852
```

Which correlates to [STREAM_NOT_STARTED](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/532178bbd4d2e6e6511fa8ffa62a15dba58c02f0/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L189).

*Description of changes:*
* In kvs_gstreamer_sample, move `putEventMetadata` after `putFrame`. When uploading a file with only a single keyframe, and if the keyframe is the first frame in the file, the event metadata will not get added. This is because the stream is started in the putFrame call, and we attempt to add the event metadata before the putFrame call. This causes the put event metadata to fail the first time, but succeed the rest of the times. Since there isn't a second keyframe in the file, the metadata would never get put.
* Also, translate the error codes into the hex format.

### DIFF
--- a/samples/kvs_gstreamer_sample.cpp
+++ b/samples/kvs_gstreamer_sample.cpp
@@ -401,12 +401,13 @@ static GstFlowReturn on_new_sample(GstElement *sink, CustomData *data) {
         if (!gst_buffer_map(buffer, &info, GST_MAP_READ)){
             goto CleanUp;
         }
-        if (CHECK_FRAME_FLAG_KEY_FRAME(kinesis_video_flags)) {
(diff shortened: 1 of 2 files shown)


unicornss (approved):

sirknightj:
Note: Going to `master` as the putEventMetadata feature is not in `develop`.

PR #1020, opened 2023-06-05
Enable caching by default

disa6302:
*Issue #, if available:*

*Description of changes:*

- Enable caching by default and set policy to ALL
- Tested kvssink and all samples in the samples folder
- Introduced new helper APIs to set this up
- If stream is not pre-created and policy is set to API_CALL_CACHE_TYPE_ENDPOINT_ONLY, the sample would encounter STREAM_NOT_FOUND error since we cache only the endpoint and not the describe call result.
- Couple of other compiler warning fixes

### DIFF
--- a/src/gstreamer/KvsSinkStreamCallbackProvider.cpp
+++ b/src/gstreamer/KvsSinkStreamCallbackProvider.cpp
@@ -71,8 +71,12 @@ STATUS
 KvsSinkStreamCallbackProvider::streamClosedHandler(UINT64 custom_data,
                                                    STREAM_HANDLE stream_handle,
                                                    UPLOAD_HANDLE upload_handle) {
-    UNUSED_PARAM(custom_data);
-    LOG_DEBUG("Reported streamClosed callback for stream handle " << stream_handle << ". Upload handle " << upload_handle);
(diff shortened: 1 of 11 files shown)


src/gstreamer/KvsSinkStreamCallbackProvider.cpp
```diff
@@ -71,8 +71,12 @@ STATUS
 KvsSinkStreamCallbackProvider::streamClosedHandler(UINT64 custom_data,
                                                    STREAM_HANDLE stream_handle,
                                                    UPLOAD_HANDLE upload_handle) {
-    UNUSED_PARAM(custom_data);
-    LOG_DEBUG("Reported streamClosed callback for stream handle " << stream_handle << ". Upload handle " << upload_handle);
```
jdelapla, line 75: Is the upload_handle irrelevant debug information?

disa6302: Yeah. It does not give us any useful details. I can add it back though. I wanted to ensure stream name was getting logged because that is descriptive

jdelapla (approved):

PR #1023, opened 2023-06-13
Invoke stopSync in NULL state transition

disa6302:
*Issue #, if available:*

*Description of changes:*
In case the streaming needs to be stopped abruptly, be it because of an error or to not let it run till EOS is received, we want to ensure stopStreamSync() is invoked when pipeline is transition to NULL after stopping data collection from buffer. Also added some logging around cleanup.

### DIFF
--- a/src/gstreamer/gstkvssink.h
+++ b/src/gstreamer/gstkvssink.h
@@ -134,6 +134,7 @@ struct _GstKvsSink {
     guint                       num_audio_streams;
     guint                       num_video_streams;
 
+
     std::unique_ptr<Credentials> credentials_;
     std::shared_ptr<KvsSinkCustomData> data;
 };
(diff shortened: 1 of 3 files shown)


src/gstreamer/gstkvssink.h
```diff
@@ -174,6 +176,7 @@ struct _KvsSinkCustomData {
     bool get_metrics;
     uint32_t frame_count;
     bool on_first_frame;
+    bool streamingStopped;
```
jdelapla, line 179: Should this be atomic? I'm not sure but I suspect gst_kvs_sink_change_state() and gst_kvs_sink_handle_buffer() aren't guaranteed to be called from the same thread.

disa6302: In the stop path, I see it being on the same thread from the logs, but yes, I have made it atomic to be on the safe side, makes sense as well.

jdelapla (approved):

PR #1039, opened 2023-08-03
Fix the config_file path for use as subdirectory

joshuafriedrick:
*Description of changes:*

If another CMake project uses KinesisVideoProducerCpp by including amazon-kinesis-video-streams-producer-sdk-cpp through an `add_subdirectory` call, then using CMAKE_SOURCE_DIR points to an invalid path. Since `ROOT` is set in the root CMakeLists.txt, we can use that variable to help point to the correct path both when building the project on its own, as well when the project is included in a parent CMake project.

This simple change allows a parent CMakeLists.txt to add this project and build it successfully.


sirknightj:
Hi @joshuafriedrick, thanks for your contribution. Could you please change the target branch to `develop`? We'll merge it in to `master` as part of the release cycle (ETA later this month).

joshuafriedrick:
Looks like this was already resolved on the `develop` branch in this commit: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/commit/3dd4d312a7423820a7670717736c822587772b27

Sorry for the noise @sirknightj !

PR #1046, opened 2023-08-15
Auth segfault fix

disa6302:
Description of changes:

* What was changed?
Added a check to cover empty session token scenario and added checks for return code to fail gracefully. The SDK does not allow empty credentials.

* Why was it changed?
To ensure correct behavior. Without the change, the samples segfault since we move on to the next steps without AwsCredentials object in producer C being created

Testing:
* Local testing of kvs gstreamer pipeline samples to ensure it runs clean
* Added dedicated unit test for empty credential combinations


src/Auth.cpp
```diff
@@ -90,6 +90,7 @@ DeviceCertToTokenFunc CredentialProvider::deviceCertToTokenCallback() {
 
 STATUS CredentialProvider::getStreamingTokenHandler(UINT64 custom_data, PCHAR stream_name, STREAM_ACCESS_MODE access_mode, PServiceCallContext p_service_call_context) {
     LOG_DEBUG("getStreamingTokenHandler invoked");
+    STATUS status = STATUS_SUCCESS;
```
sirknightj, line 93: The initial value doesn't matter since it will get overwritten on L115 and L120. There is no Cleanup clause for this method

sirknightj (approved):

src/Auth.cpp
```diff
@@ -111,14 +112,25 @@ STATUS CredentialProvider::getStreamingTokenHandler(UINT64 custom_data, PCHAR st
     freeAwsCredentials(&this_obj->security_token_);
 
     // Store the buffer so we can release it at the end
-    createAwsCredentials((PCHAR) access_key.c_str(), access_key_len, (PCHAR) secret_key.c_str(), secret_key_len,
-                         (PCHAR) session_token.c_str(), session_token_len, expiration, &this_obj->security_token_);
+    if(IS_EMPTY_STRING(session_token.c_str())) {
```
sirknightj, line 115: Q: is 0 length access key secret key handled here or is that handled somewhere else

src/Auth.cpp
disa6302: That is ok. It is always preferable to set an initial value for variable declarations. Never leave to assumptions :)

src/Auth.cpp
disa6302: 0 is an allowed value. The API calculates the length anyways if it is supplied as 0.

sirknightj (dismissed):

sirknightj (approved):
LGTM!

PR #1047, opened 2023-08-17
fix(kvs/sink_gstreamer_sample): Use videotestsrc and add additional debug logging

sirknightj:
### Issue #, if available
When running the sample on a mac in live_source mode (`./kvs_gstreamer_sample demo-stream`), we would get "Elements could not be linked." error. We tested this on both m1 mac, and intel mac, each using `GStreamer Core Library version 1.22.5` and encountered this error both times.

### What was changed, and how was it changed?
* In live_source mode, use `videotestsrc` instead of `autovideosrc`
* Add additional debug logs
* The same was also applied to `kvssink_gstreamer_sample`, as these samples are very similar.
  * Fix typo in `kvssink_gstreamer_sample`: `gracefully`.

### Why was it changed?
* We determined that `data->h264_stream_supported` is `true`, therefore, [this pipeline](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L845) was getting created.
* `autovideosrc` has [ANY pads](https://gstreamer.freedesktop.org/documentation/autodetect/autovideosrc.html?gi-language=c#src), so, when [data.h264_stream_supported being determined](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L448-L450), [the intersection](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L432) will always return `true`. `autovideosrc`'s pads do not accurately reflect the available elements on the device, as that is determined during the application runtime. Both this [if statement](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L447C26-L447C26) and [this if statement](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L449) return `true` with `autovideosrc`.
* Neither `autovideosrc`, `videotestsrc` or [`avfvideosrc`](https://gstreamer.freedesktop.org/documentation/applemedia/avfvideosrc.html?gi-language=c#src) output `video/x-h264`, therefore it cannot be directly linked to [`h264parse`](https://gstreamer.freedesktop.org/documentation/videoparsersbad/h264parse.html?gi-language=c#src).
  * You can confirm this by using the following: `gst-launch-1.0 <src element> ! capsfilter name=filter caps="video/x-h264" ! fakesink` and seeing "`WARNING: erroneous pipeline: could not link <src element> to filter`". The application should be choosing to construct the [top pipeline with the encoder element](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/4c4afff63f3371a69be2472bb0bee654ef09abd3/samples/kvs_gstreamer_sample.cpp#L836-L842), rather than [the bottom pipeline](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/4c4afff63f3371a69be2472bb0bee654ef09abd3/samples/kvs_gstreamer_sample.cpp#L844-L849).
* By switching `autovideosrc` to `videotestsrc`, the correct caps are advertised, and the sample application (using `gst_caps_can_intersect`) is able to choose the correct pipeline to construct.

### Testing
* Ran both samples in live source mode: `./kvs_gstreamer_sample demo-stream` and `./kvssink_gstreamer_sample demo-stream` and verified playback in the AWS management console.

image

### DIFF
--- a/samples/kvs_gstreamer_sample.cpp
+++ b/samples/kvs_gstreamer_sample.cpp
@@ -702,6 +702,7 @@ int gstreamer_live_source_init(int argc, char* argv[], CustomData *data, GstElem
             LOG_DEBUG("Using videotestsrc");
         } else {
             LOG_ERROR("Failed to create videotestsrc");
+            return 1;
         }
     } else {
         // Failed creating vtenc - check pi hardware encoder
@@ -717,17 +718,20 @@ int gstreamer_live_source_init(int argc, char* argv[], CustomData *data, GstElem
                 LOG_DEBUG("Using x264enc");
             } else {
                 LOG_ERROR("Failed to create x264enc");
+                return 1;
             }
         }
         source = gst_element_factory_make("v4l2src", "source");
         if (source) {
             LOG_DEBUG("Using v4l2src");
         } else {
+            LOG_DEBUG("Failed to create v4l2src, trying ksvideosrc")
             source = gst_element_factory_make("ksvideosrc", "source");
             if (source) {
                 LOG_DEBUG("Using ksvideosrc");
             } else {
-                LOG_ERROR("Unable to create src element. Tried v4l2src and ksvideosrc");
+                LOG_ERROR("Failed to create ksvideosrc");
+                return 1;
             }
         }
         vtenc = false;
--- a/samples/kvs_gstreamer_sample.cpp
+++ b/samples/kvs_gstreamer_sample.cpp
@@ -1158,7 +1158,10 @@ int main(int argc, char* argv[]) {
                 LOG_DEBUG("Attempt to upload file: " << data.file_list[i].path);
 
                 // control will return after gstreamer_init after file eos or any GST_ERROR was put on the bus.
-                gstreamer_init(argc, argv, &data);
+                if (gstreamer_init(argc, argv, &data) != 0) {
+                    LOG_ERROR("Failed to initialize gstreamer");
+                    return 1;
+                }
 
                 // check if any stream error occurred.
                 stream_status = data.stream_status.load();
@@ -1208,7 +1211,10 @@ int main(int argc, char* argv[]) {
 
     } else {
         // non file uploading scenario
-        gstreamer_init(argc, argv, &data);
+        if (gstreamer_init(argc, argv, &data) != 0) {
+            LOG_ERROR("Failed to initialize gstreamer");
+            return 1;
+        }
         if (STATUS_SUCCEEDED(stream_status)) {
             // if stream_status is success after eos, send out remaining frames.
             data.kinesis_video_stream->stopSync();
--- a/samples/kvs_gstreamer_sample.cpp
+++ b/samples/kvs_gstreamer_sample.cpp
@@ -441,8 +441,10 @@ static bool format_supported_by_source(GstCaps *src_caps, GstCaps *query_caps, i
 static bool resolution_supported(GstCaps *src_caps, GstCaps *query_caps_raw, GstCaps *query_caps_h264,
                                  CustomData &data, int width, int height, int framerate) {
     if (query_caps_h264 && format_supported_by_source(src_caps, query_caps_h264, width, height, framerate)) {
+        LOG_DEBUG("src supports h264")
         data.h264_stream_supported = true;
     } else if (query_caps_raw && format_supported_by_source(src_caps, query_caps_raw, width, height, framerate)) {
+        LOG_DEBUG("src supports raw")
         data.h264_stream_supported = false;
     } else {
         return false;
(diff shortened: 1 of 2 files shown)


samples/kvs_gstreamer_sample.cpp
```diff
@@ -707,7 +710,10 @@ int gstreamer_live_source_init(int argc, char* argv[], CustomData *data, GstElem
             isOnRpi = false;
         }
         source = gst_element_factory_make("v4l2src", "source");
-        if (!source) {
+        if (source) {
+            LOG_DEBUG("Using v4l2src");
+        } else {
+            LOG_DEBUG("Using ksvideosrc");
             source = gst_element_factory_make("ksvideosrc", "source");
```
hassanctech, line 717: Hmm I know this was always there but this looks Windows Specific [https://gstreamer.freedesktop.org/documentation/winks/index.html?gi-language=c]... so basically we're assuming if none of the other sources succeeded then it's Windows.  That's not good.  At the very least we should have a nul check here as well and log that it failed.  I understand in the  lines below we will pint error Not all elements could be created.  But I think it's better to explicitly log if source is still null that this element specifically couldn't be created.

samples/kvs_gstreamer_sample.cpp
```diff
@@ -694,7 +696,8 @@ int gstreamer_live_source_init(int argc, char* argv[], CustomData *data, GstElem
     // Attempt to create vtenc encoder
     encoder = gst_element_factory_make("vtenc_h264_hw", "encoder");
     if (encoder) {
-        source = gst_element_factory_make("autovideosrc", "source");
+        LOG_DEBUG("Using videotestsrc")
```
hassanctech, line 699: This log is misleading so it's only using `videotestsrc` if we were able to successfully create source and it's not null..  we're just trying `vidoetestsrc` here.  Further I would move up setting vtenc = true to the first line inside the if(encoder) { section since if encoder is TRUE we know we've successfully created it.  Shouldn't really change logic much but I think the logic flow is better reading the code otherwise it kind of reads like vtenc is being set as a result  of creating videotestsrc.

samples/kvs_gstreamer_sample.cpp
sirknightj: Added

samples/kvs_gstreamer_sample.cpp
sirknightj: Sure thing. Done.

samples/kvs_gstreamer_sample.cpp
```diff
@@ -696,25 +696,39 @@ int gstreamer_live_source_init(int argc, char* argv[], CustomData *data, GstElem
...
+            } else {
+                LOG_ERROR("Failed to create x264enc");
+            }
         }
         source = gst_element_factory_make("v4l2src", "source");
         if (source) {
             LOG_DEBUG("Using v4l2src");
         } else {
-            LOG_DEBUG("Using ksvideosrc");
             source = gst_element_factory_make("ksvideosrc", "source");
```
hassanctech, line 726: Since this is a sample it would be better if we could communicate exactly which one.  So how about a LOG line stating unable to create `v4l2src` and then the last one can only say it was unable to create `ksvideosrc` instead of combining them.

in the if(vtenc) block if we fail to create `videotestsrc` that is a fatal error we can terminate the application there.  in the else block we attempt a few so I would check after the else {} section is complete to see if we have a valid src if not, terminate the application.  I'm not a fan of having an if (!a || !b || !c ||....) thing and the customer now has to figure out well which one was it that failed.

sirknightj: Sure, we can go with that.

sirknightj: If we fail to create videotestsrc in the if(vtenc) block, we will still exit on the next line below. We can move the check upwards if needed. The downside is that the exit condition is messier.

hassanctech (commented):
Having `gstreamer_life_source_init` return 1 to indicate 1 is fine.  However I tried the code and found that this function is called inside `gstreamer_init` which is in turn called usually in main and in most cases it is called without checking the return value.  Can we please correct all of those as well so we do not have silent failures.

https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L1119

https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/e8125ee363c4e77e11b15398a8b77f4e8a5791a5/samples/kvs_gstreamer_sample.cpp#L1169

sirknightj:
> Can we please correct all of those as well so we do not have silent failures.

Added.

hassanctech (approved):
lgtm

PR #1078, opened 2023-10-03
[DRAFT] Put frame return error codes support start stop

hassanctech:
*Issue #, if available:*

*Description of changes:*

The `putKinesisVideoFrame` API in PIC returns a detailed `STATUS` code.  Unfortunately in Producer CPP the `KinesisVideoStream` class has a `putFrame` member function which unfortunately swallows the return STATUS code from the lower level PIC API and only returns a boolean.  Further in the current implementation of `kvssink` GStreamer element the boolean return code is completely ignored so we do not do anything at all when the putFrame call failed with some error.  In addition we do not properly handle GStreamer pipeline state transitions in the `kvssink` GStreamer element, in particular the underlying Producer SDK supports stopping a stream and then restreaming.  This requires support for more transitions inside the `gst_kvs_sink_change_state` function, specifically more case statements need to be added to support different types of `GstStateChange`, see [ref](https://api.gtkd.org/gstreamer.c.types.GstStateChange.html) for details about possible state transitions in a `GStreamer` pipeline.

- [X] `KinesisVideoStream::putFrame` should return a `STATUS`
- [X] `kvssink` GStreamer elements should not ignore the return status from the putFrame call.  When an error is received from `putFrame` a message should be placed on the GStreamer bus indicating 
- [ ] Complete support for different values of `GstStateChange` in `gst_kvs_sink_change_state`
- [ ] Explicit handler for receipt of EOS [need to validate if this is actually necessary]
- [ ] Test start / stop of pipeline transition between different states.  
- [ ] Test or Sample update to show how to use the new messages we put on the bus and how an application can properly respond to them.

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1211,13 +1211,16 @@ void create_kinesis_video_frame(Frame *frame, const nanoseconds &pts, const nano
     frame->trackId = static_cast<UINT64>(track_id);
 }
 
-bool put_frame(shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len, const nanoseconds &pts,
+STATUS
+put_frame(std::shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len, const nanoseconds &pts,
           const nanoseconds &dts, FRAME_FLAGS flags, uint64_t track_id, uint32_t index) {
 
+    STATUS put_frame_status = STATUS_SUCCESS;
     Frame frame;
+
     create_kinesis_video_frame(&frame, pts, dts, flags, frame_data, len, track_id, index);
-    bool ret = data->kinesis_video_stream->putFrame(frame);
-    if(data->get_metrics && ret) {
+    put_frame_status = data->kinesis_video_stream->putFrame(frame);
+    if(data->get_metrics && STATUS_SUCCEEDED(put_frame_status)) {
         if(CHECK_FRAME_FLAG_KEY_FRAME(flags)  || data->on_first_frame){
             KvsSinkMetric *kvs_sink_metric = new KvsSinkMetric();
             kvs_sink_metric->stream_metrics = data->kinesis_video_stream->getMetrics();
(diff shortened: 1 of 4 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -1374,6 +1378,11 @@ gst_kvs_sink_handle_buffer (GstCollectPads * pads,
         gst_buffer_unref (buf);
     }
 
+    if (STATUS_FAILED(put_frame_status)) {
+        GST_ELEMENT_WARNING (kvssink, STREAM, FAILED, (NULL),
```
stefankiesz, line 1382: Can we consider changing this to:

<pre style="color:blue;">
GST_ELEMENT_WARNING (kvssink, <b>RESOURCE</b>, <b>GST_RESOURCE_ERROR_WRITE,</b> (NULL),
</pre>

Based on these descriptions in GStreamer documentation:
> **Stream** errors are for anything related to the stream being processed: format errors, media type errors, ... They're typically used by decoders, demuxers, converters, ...
[here](https://gstreamer.freedesktop.org/documentation/gstreamer/gsterror.html?gi-language=c#GstStreamError:~:text=Stream%20errors%20are%20for%20anything%20related%20to%20the%20stream%20being%20processed%3A%20format%20errors%2C%20media%20type%20errors%2C%20...%20They%27re%20typically%20used%20by%20decoders%2C%20demuxers%2C%20converters%2C%20...)

> **Resource** errors are for any resource used by an element: memory, files, network connections, process space, ... They're typically used by source and sink elements.
[here](https://gstreamer.freedesktop.org/documentation/gstreamer/gsterror.html?gi-language=c#GstResourceError:~:text=Resource%20errors%20are%20for%20any%20resource%20used%20by%20an%20element%3A%20memory%2C%20files%2C%20network%20connections%2C%20process%20space%2C%20...%20They%27re%20typically%20used%20by%20source%20and%20sink%20elements.)

stefankiesz:
I rebased on top of develop and updated CI tests to handle STATUS type returns in [this](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1119) PR.

Issue #1068, opened 2023-09-21
[BUG] kvssink build on Alpine linux fails with segfault 


alan-willard-simplisafe:
**Logging**
##### Output when running `gst-launch-1.0` with the `kvssink`
>/opt/amazon-kinesis-video-streams-producer-sdk-cpp/build # gst-launch-1.0 -v videotestsrc ! x264enc ! kvssink stream-name=test aws-region=us-east-1 access-key=test secret-key=test
Setting pipeline to PAUSED ...
[INFO ] [21-09-2023 15:42:31:208.267 UTC] createKinesisVideoClient(): Creating Kinesis Video Client
[INFO ] [21-09-2023 15:42:31:208.363 UTC] heapInitialize(): Initializing native heap with limit size 134217728, spill ratio 0% and flags 0x00000001
[INFO ] [21-09-2023 15:42:31:208.383 UTC] heapInitialize(): Creating AIV heap.
[INFO ] [21-09-2023 15:42:31:208.414 UTC] heapInitialize(): Heap is initialized OK
[DEBUG] [21-09-2023 15:42:31:208.565 UTC] getSecurityTokenHandler invoked
[DEBUG] [21-09-2023 15:42:31:208.602 UTC] Refreshing credentials. Force refreshing: 0 Now time is: 1695310951208599221 Expiration: 0
[INFO ] [21-09-2023 15:42:31:208.632 UTC] createDeviceResultEvent(): Create device result event.
[DEBUG] [21-09-2023 15:42:31:208.645 UTC] clientReadyHandler invoked
[INFO ] [21-09-2023 15:42:31:208.718 UTC] try creating stream
[INFO ] [21-09-2023 15:42:31:208.761 UTC] Creating Kinesis Video Stream test
[INFO ] [21-09-2023 15:42:31:208.791 UTC] createKinesisVideoStream(): Creating Kinesis Video Stream.
[INFO ] [21-09-2023 15:42:31:208.841 UTC] logStreamInfo(): SDK version: 0cfcd36648ced65a1cf740bcd9f46034ef8fff50
[DEBUG] [21-09-2023 15:42:31:208.859 UTC] logStreamInfo(): Kinesis Video Stream Info
[DEBUG] [21-09-2023 15:42:31:208.868 UTC] logStreamInfo(): 	Stream name: test
[...]

PR #1084, opened 2023-10-11
Alpine stackoverflow fix

jdelapla:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/1068

*Description of changes:*
Allow option to change thread stack size to support Alpine


CMakeLists.txt
```diff
@@ -42,6 +43,10 @@ else()
   set(LINKAGE SHARED)
 endif()
 
+if(CONSTRAINED_DEVICE)
```
disa6302, line 46: We might not need this for CPP actually. It should be enough to pass in this value in producer C cmake file right as a cmake build arg? Something like this:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt#L11-L20

So when CPP is run with `cmake .. -DCONSTRAINED_DEVICE=ON`, producer C would add the definition. 
Curious if we tried this.

jdelapla: I had expected that as well, but when I tested it it did not work. Perhaps I made a mistake, I'll repeat the test.

jdelapla: You're correct, I had misspelled the flag.

disa6302 (approved):

PR #1095, opened 2023-10-28
Add clang formatting script and format all smaple and src files

stefankiesz:
Noticed there was a formatting error in the kvssink sample, brought in CLANG formatting used in C SDK to prevent this in future.

<br>


tst/ProducerApiTest.cpp
```diff
@@ -298,15 +279,10 @@ TEST_F(ProducerApiTest, create_produce_start_stop_stream)
...
-                                                   << " Id: " << frame.index
+            LOG_DEBUG("Putting frame for stream: " << kinesis_video_stream->getStreamName() << strstrm.str() << " Id: " << frame.index
                                                    << ", Key Frame: "
-                                                   << (((frame.flags & FRAME_FLAG_KEY_FRAME) == FRAME_FLAG_KEY_FRAME)
-                                                       ? "true" : "false")
-                                                   << ", Size: " << frame.size
-                                                   << ", Dts: " << frame.decodingTs
-                                                   << ", Pts: " << frame.presentationTs);
+                                                   << (((frame.flags & FRAME_FLAG_KEY_FRAME) == FRAME_FLAG_KEY_FRAME) ? "true" : "false")
+                                                   << ", Size: " << frame.size << ", Dts: " << frame.decodingTs << ", Pts: " << frame.presentationTs);
```
hassanctech, line 285: I think we should exclude the tests from the clang format check, the version before here looks better than how it formatted it.

src/gstreamer/gstkvssink.h
```diff
@@ -70,111 +65,97 @@ typedef struct _KvsSinkMetric KvsSinkMetric;
...
-
-    guint                       num_streams;
-    guint                       num_audio_streams;
-    guint                       num_video_streams;
-
+    gchar* stream_name;
+    gchar* user_agent;
+    guint retention_period_hours;
+    gchar* kms_key_id;
+    STREAMING_TYPE streaming_type;
```
hassanctech, line 91: Is there a way to keep it how it was before, I think the before version here is clearer. We should stick to one way and be consistent, the consistency is most important.

stefankiesz: In Producer C, we do use this same style of not having indentations to align variable names - there are indentations for "#define"s though. I tested and searched around to find that the way clang-format works is it parses all the code and then puts it back together according to the configured format, so there is no way to remove this formatting. There may be a formatting rule that would add this spacing between variable type and name, but it would then apply to _all_ variables. The two solutions I see are to either use another formatter, or place the following around code we do not want to format:
```
// clang-format off
...
// clang-format on
```

niyatim23 (commented):
If we introduce the files for formatting, and add a build in the CI but don't update our codebase to reflect the format, the check in the CI will fail. Is there any specific reason you removed the formatting from all the files you had added previously?

stefankiesz:
@niyatim23 I was testing whether it would affect the failing address sanitizer CI job. Just reverted those commits.

PR #1105, opened 2023-11-18
Fix GCC and Windows CI Tests

stefankiesz:
### Mac GCC CI Test ###
Changed to using a direct and versioned compiler path - just pointing to "gcc" and "g++" will map to CLANG compiler on Mac.

<br>

### Windows MSVC CI Test ###
Updated PKG_CONFIG_EXECUTABLE to reflect current Chocolatey GSTreamer install path - it is  in D: drive now, not C: drive.

<br>
<br>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ else()
 endif()
 
 if (WIN32)
-  set(PKG_CONFIG_EXECUTABLE "C:\\gstreamer\\1.0\\x86_64\\lib\\pkg-config.exe")
+  set(PKG_CONFIG_EXECUTABLE "C:\\gstreamer\\1.0\\msvc_x86_64\\lib\\pkg-config.exe")
 endif()
 
 ############# Enable Sanitizers ############
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,10 +126,6 @@ else()
   set(Log4cplus ${LOG4CPLUS_LIBRARIES})
 endif()
 
-if (WIN32)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ else()
 endif()
 
 if (WIN32)
-  set(PKG_CONFIG_EXECUTABLE "C:\\gstreamer\\1.0\\x86_64\\bin\\pkg-config.exe")
+  set(PKG_CONFIG_EXECUTABLE "D:\\gstreamer\\1.0\\x86_64\\bin\\pkg-config.exe")
```
niyatim23, line 130: If this path is specific to the CI, should we make allowing customizing it? What was this path on your local Windows machine (if you were able to try this)?

stefankiesz: On local machine it is C:, I think best way to handle this is to specify the location for Chocolatey installs so we can use C: for everything, unless customers are finding that it sometimes does get directed to D: for them as well. Going to look into this now.

niyatim23: Okay, we can keep it D:\. Let's a note in the readme for this, mentioning that the customers can change this if this path is different in their installation

stefankiesz: I tried setting gstreamer installation path using by passing an install argument to choco command, and that resulted in neither the C nor D drive having the gstreamer installation. I ended up putting a check in CMake to see if the C drive path exists, else use D drive.

niyatim23 (approved):
Nice work on fixing the both!

PR #1113, opened 2023-11-29
Fix KVSSINK State Transition Deadlock

stefankiesz:
Move parent state transition to be in between upward and downward element transitions as per [GStreamer documentation](https://gstreamer.freedesktop.org/documentation/plugin-development/basics/states.html?gi-language=c#:~:text=Note%20that%20upwards,destroying%20allocated%20resources.).

Tested using `kvssink_gstreamer_sample` to successfully ingest stream and cleanly exit upon sigint with videotestsrc, filesrc, and rtspsrc sources.

<br>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   mac-os-build-clang:
-    runs-on: macos-13
+    runs-on: macos-12
     env:
       AWS_KVS_LOG_LEVEL: 2
       CC: /usr/bin/clang
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   mac-os-build-clang:
-    runs-on: macos-12
+    runs-on: macos-13
     env:
       AWS_KVS_LOG_LEVEL: 2
       CC: /usr/bin/clang
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1572,12 +1572,25 @@ init_track_data(GstKvsSink *kvssink) {
 
 static GstStateChangeReturn
 gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
+    /*
+        The below state transition cases are separated into two switch blocks:
+        one for upward (NULL->READY->PAUSED->PLAYING) transitions and one for
+        downward (PLAYING->PAUSED->READY->NULL) transitions. It is necessary to transition
+        an element's parent class state after any of the element's upward transitions but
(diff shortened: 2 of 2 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -1609,6 +1609,13 @@ gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
         case GST_STATE_CHANGE_READY_TO_PAUSED:
             gst_collect_pads_start (kvssink->collect);
             break;
+        default:
+            break;
+    }
+
+    ret = GST_ELEMENT_CLASS (parent_class)->change_state(element, transition);
```
niyatim23, line 1616:
```suggestion
    ret = GST_ELEMENT_CLASS (parent_class)->change_state(element, transition);
    if (ret == GST_STATE_CHANGE_FAILURE) {
        goto CleanUp;
    }
```

stefankiesz:
@hassanctech No errors from my testing, we have a customer who states they had a deadlock before implementing this change. Let's discus setting up a test, maybe based on their implementation.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   mac-os-build-clang:
-    runs-on: macos-12
+    runs-on: macos-13
```
niyatim23, line 14: Let us change this back to 12, I have a fix for the build on 13. I can address it in another PR later

stefankiesz: Done

src/gstreamer/gstkvssink.cpp
```diff
@@ -1609,6 +1609,16 @@ gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
         case GST_STATE_CHANGE_READY_TO_PAUSED:
             gst_collect_pads_start (kvssink->collect);
             break;
+        default:
```
hassanctech, line 1612: This is great, at the start of this function can you add a comment block explaining the change, so it's visible to someone browsing the code itself, in addition to the PR where I see it's already added.

stefankiesz: Done!

hassanctech (approved):
nice!

src/gstreamer/gstkvssink.cpp
```diff
@@ -1609,6 +1622,18 @@ gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
...
+            break;
+    }
+
+    // Parent class transition
+    ret = GST_ELEMENT_CLASS (parent_class)->change_state(element, transition);
```
unicornss, line 1630: suggest adding  the return status and also the (parent) element and transition as debug logs..

PR #1116, opened 2023-12-05
Verifying versioned libs

jdelapla:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/723

*Description of changes:*
Versioning libraries


CMakeLists.txt
```diff
@@ -8,6 +8,16 @@ project(KinesisVideoProducerCpp VERSION 3.4.0)
 set(CMAKE_CXX_STANDARD 11)
 include(GNUInstallDirs)
 
+set(KINESIS_VIDEO_PRODUCER_CPP_MAJOR_VERSION 3)
+set(KINESIS_VIDEO_PRODUCER_CPP_MINOR_VERSION 1)
```
disa6302, line 12: We should change this to 4 given our current version is 3.4.1

CMakeLists.txt
```diff
@@ -8,6 +8,16 @@ project(KinesisVideoProducerCpp VERSION 3.4.0)
...
+set(KINESIS_VIDEO_PRODUCER_CPP_PATCH_VERSION 1)
+set(KINESIS_VIDEO_PRODUCER_CPP_VERSION ${KINESIS_VIDEO_PRODUCER_CPP_MAJOR_VERSION}.${KINESIS_VIDEO_PRODUCER_CPP_MINOR_VERSION}.${KINESIS_VIDEO_PRODUCER_CPP_PATCH_VERSION})
+
+set(KINESIS_VIDEO_PRODUCER_CPP_MAJOR_VERSION 3)
+set(KINESIS_VIDEO_PRODUCER_CPP_MINOR_VERSION 1)
```
disa6302, line 17: Same here.

CMakeLists.txt
hassanctech: +1

CMakeLists.txt
sirknightj: What's the reason to have this twice?

sirknightj:
I also updated the property set for the user agent to use this new value

CMakeLists.txt
sirknightj: In the original, this was only there once. So I will remove the extra one.

disa6302 (approved):

PR #1122, opened 2024-01-02
[FEATURE] Support putFragmentMetadata from kvssink

niyatim23:
Issue #, if available:
https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/discussions/1031

*What was changed?*
Modify `kvssink` and sample application to support adding metadata per fragment

*Why was it changed?*
This feature was missing from `kvssink` and was requested by multiple customers

*How was it changed?*
- This feature allows adding multiple metadata tags to fragments until a limit (10 metadata per fragment) is reached. To repeatedly put the metadata while streaming, a timer has been used which triggers a function `put-metadata` every 2 seconds. This application-level logic can be customized by the customers to be fired on their own set of events / frequency. `put-metadata` can be modified by the customers to generate metadata tags to be put. Currently, they consume values from a global instance of CustomData whose counter is updated within `put-metadata` itself. `put-metadata` also calls `put-fragment-metadata`
- `put-fragment-metadata` generates the `GstStructure` expected by the `GST_EVENT_CUSTOM_DOWNSTREAM` and ultimately triggers the `GST_EVENT_CUSTOM_DOWNSTREAM` that handles the tags and calls the underlying KVS API `putFragmentMetadata`. 
- When this event is triggered, it checks whether the structure is in the expected format. If it has, it warns the customer about it through the logs without disrupting the streaming. 

Tested the following locally:
- Tested that the metadata was persisted in the video by checking the video dump with `mkvinfo`
- Tested that the metadata marked as `persistent` is indeed persisted for each fragment of the video
- Tested that the metadata marked as `non-persistent` is not persisted for each fragment of the video
- Tested that the metadata that is to be put, toggles between `persistent` and `non-persistent` and the both of them are marked correctly in the video
- Tested that the streaming continues despite failures in the putFragmentMetadata
- Tested that sending empty values for persistent metadata removes it
- `mkvinfo` on a video dump from the tests (even metadata-keys are persistent, odd ones are non-persistent): [mkvinfo-metadata.txt](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/files/13842435/mkvinfo-metadata.txt)


src/gstreamer/gstkvssink.cpp
```diff
@@ -1217,14 +1224,16 @@ bool put_frame(shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len,
     Frame frame;
     create_kinesis_video_frame(&frame, pts, dts, flags, frame_data, len, track_id, index);
     bool ret = data->kinesis_video_stream->putFrame(frame);
-    if(data->get_metrics && ret) {
-        if(CHECK_FRAME_FLAG_KEY_FRAME(flags)  || data->on_first_frame){
+        
+    if(CHECK_FRAME_FLAG_KEY_FRAME(flags)  || data->on_first_frame) {
+        data->fragment_metadata_count = data->persisted_fragment_metadata_count;
```
stefankiesz, line 1229: Can you help me to understand why we don't set `fragment_metadata_count` to zero here? Seems like this would need to be reset to zero at the start of every fragment to not exceed the max tags per fragment.

And if `persisted_fragment_metadata_count` is a count of the total tags persisted throughout the whole stream, can we add a `total_` prefix to this variable name to make that more clear?

niyatim23: Persisted tags by default means that we are going to add them to all the subsequent fragments so to respect the limit, it has been initialized to persisted_fragment_metadata_count. I did not include "total" in the name as "persisted", here, means that we are going to add it to every fragment from now on

src/gstreamer/gstkvssink.cpp
```diff
@@ -1157,15 +1153,18 @@ gst_kvs_sink_handle_sink_event (GstCollectPads *pads,
             gboolean persistent;
             bool is_persist;
 
-            if (!gst_structure_has_name(structure, KVS_ADD_METADATA_G_STRUCT_NAME)) {
+            if (!gst_structure_has_name(structure, KVS_ADD_METADATA_G_STRUCT_NAME) || 
+                    data->fragment_metadata_count >= MAX_FRAGMENT_METADATA_TAGS) {
+                ret = FALSE;
+                LOG_WARN("Current fragment's metadata count  " << data->fragment_metadata_count << ". Max limit reached. Current metadata cannot be persisted.");
```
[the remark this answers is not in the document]
niyatim23: Moved the structure check along with other similar checks

src/gstreamer/gstkvssink.cpp
stefankiesz: Ah got it. Can we consider renaming this to `persistent_` rather than `persisted_` in that case? And do you think we should also provide an upstream message type to remove a persistent tag?

src/gstreamer/gstkvssink.cpp
```diff
@@ -1157,27 +1153,39 @@ gst_kvs_sink_handle_sink_event (GstCollectPads *pads,
             gboolean persistent;
             bool is_persist;
```
stefankiesz, line 1154: (nit) Can this be `is_persistent`

src/gstreamer/gstkvssink.cpp
niyatim23: I don't think we have an API to remove the flag from persisted metadata actually. I named it "persisted" as they have been persisted into our media already in the previous calls

hassanctech: As per: https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/how-meta.html#how-meta-add

The way to "remove" the metadata is to call the API with an empty *value* will remove the metadata from the queue, we should support this logic as well.

niyatim23: The customer can modify the function that sets the names and values of the metadata so I don't think this will need any modification. When Stefan suggested removing, I thought he was talking about a new KVS API to do that and we don't have that but I'm sure we can call the existing APIs with different values

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -683,28 +696,34 @@ int gstreamer_init(int argc, char *argv[], CustomData *data) {
...
+    GstElement *pipeline, *kvssink;
     int ret;
     GstStateChangeReturn gst_ret;
 
     // Reset first frame pts
     data->first_pts = GST_CLOCK_TIME_NONE;
 
+    kvssink = gst_element_factory_make("kvssink", "kvssink");
+    if (!kvssink) {
+        LOG_ERROR("Failed to create kvssink");
```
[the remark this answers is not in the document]
niyatim23: That isn't in the scope of this PR, this is still work in progress

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -250,15 +254,31 @@ void determine_credentials(GstElement *kvssink, CustomData *data) {
...
+
+    metadata_name_stream << "metadata_name_" << data_global.metadata_counter;
+    metadata_value_stream << "metadata_value_" << data_global.metadata_counter;
+    
+    data_global.metadata_counter++;
+    data_global.persist_flag = !data_global.persist_flag;
+
+    if (!put_fragment_metadata(element, metadata_name_stream.str(), metadata_value_stream.str(), data_global.persist_flag)) {
+        LOG_WARN("Failed to put fragment metadata, going to stop the timer that sends the metadata");
+        g_source_remove(data_global.fragment_metadata_timer_id);
```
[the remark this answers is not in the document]
niyatim23: This is a work in progress after the recent discussions

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -718,12 +758,16 @@ int gstreamer_init(int argc, char *argv[], CustomData *data) {
     g_signal_connect(G_OBJECT(bus), "message::error", (GCallback) error_cb, data);
     g_signal_connect(G_OBJECT(bus), "message::eos", G_CALLBACK(eos_cb), data);
     gst_object_unref(bus);
+
+    // Create a GStreamer timer to generate and put fragment metadata tags every 2 seconds
+    data->fragment_metadata_timer_id = g_timeout_add_seconds(2, reinterpret_cast<GSourceFunc>(put_metadata), kvssink);
```
disa6302, line 763: Nit: #define the value 2.

src/gstreamer/gstkvssink.cpp
```diff
@@ -1231,7 +1228,6 @@ put_frame(std::shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len,
             kvs_sink_metric->client_metrics = data->kinesis_video_producer->getMetrics();
             kvs_sink_metric->frame_pts = frame.presentationTs;
             kvs_sink_metric->on_first_frame = data->on_first_frame;
-            data->on_first_frame = false;
```
disa6302, line 1234: Why are we removing this? In a weird situation where every frame is a key frame, this loop with run infinitely.

src/gstreamer/gstkvssink.h
```diff
@@ -207,4 +212,13 @@ struct _KvsSinkMetric {
     bool on_first_frame;
 };
 
+static bool inline put_fragment_metadata(GstElement* element, const std::string name, const std::string value, bool persistent) {
```
disa6302, line 215: Curious if there is a reason this is inline.

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -683,28 +693,34 @@ int gstreamer_init(int argc, char *argv[], CustomData *data) {
...
 
-    GstElement *pipeline;
+    GstElement *pipeline, *kvssink;
     int ret;
     GstStateChangeReturn gst_ret;
 
     // Reset first frame pts
     data->first_pts = GST_CLOCK_TIME_NONE;
 
+    kvssink = gst_element_factory_make("kvssink", "kvssink");
```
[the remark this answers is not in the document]
niyatim23: Because it is common to all cases and used in a common section later
`data->fragment_metadata_timer_id = g_timeout_add_seconds(2, reinterpret_cast<GSourceFunc>(put_metadata), kvssink);`

src/gstreamer/gstkvssink.cpp
niyatim23: Didn't intend to remove it, missed it while updating that section. Added it back. Thanks for pointing it out

samples/kvssink_gstreamer_sample.cpp
niyatim23: Done!

src/gstreamer/gstkvssink.h
niyatim23: I needed that at one point during the development and I assumed that I needed it later. I checked that I don't need it anymore. Removed it.

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -250,15 +270,38 @@ void determine_credentials(GstElement *kvssink, CustomData *data) {
...
+static void put_metadata(GstElement* element) {
+    std::ostringstream metadata_name_stream, metadata_value_stream;
+
+    metadata_name_stream << "metadata_name_" << data_global.metadata_counter;
+    metadata_value_stream << "metadata_value_" << data_global.metadata_counter;
+    
+    data_global.metadata_counter++;
+    data_global.persist_flag = !data_global.persist_flag;
+
+    if (!put_fragment_metadata(element, metadata_name_stream.str(), metadata_value_stream.str(), data_global.persist_flag)) {
```
sirknightj, line 295: Want to confirm the logic here. We add metadata continuously until the queue fills up with persistent wetadata and then shut this timer down after. so after some time we will have always the same set of persistent metadata being added?

niyatim23: Yes, because we wanted to make sure that toggling between the two works. Besides, this is a customizable function so the logic to generate the three parameters can be anything. I wanted to keep it simple and demonstrate that the logic works at the same time

sirknightj: It will look bad to unfamiliar users if the error/warn log line is printed during runtime.

Since this is intentional behavior, can adjust to make it more clear that it isn't a true error or the error is expected.

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -155,6 +161,10 @@ static bool resolution_supported(GstCaps *src_caps, GstCaps *query_caps_raw, Gst
 
 /* callback when eos (End of Stream) is posted on bus */
 static void eos_cb(GstElement *sink, GstMessage *message, CustomData *data) {
+    if (data->fragment_metadata_timer_id != 0) {
+        g_source_remove(data->fragment_metadata_timer_id);
```
sirknightj, line 165: By explicitly setting data->fragment_metadata_timer_id to 0, you make it clear that the timer is no longer active, which can help prevent potential issues related to accessing stale timer IDs or mistakenly assuming a timer is active when it's not.

src/gstreamer/gstkvssink.h
```diff
@@ -207,4 +212,13 @@ struct _KvsSinkMetric {
     bool on_first_frame;
 };
 
+static bool put_fragment_metadata(GstElement* element, const std::string name, const std::string value, bool persistent) {
+  GstStructure *metadata = gst_structure_new_empty(KVS_ADD_METADATA_G_STRUCT_NAME);
+  gst_structure_set(metadata, KVS_ADD_METADATA_NAME, G_TYPE_STRING, name.c_str(), 
+                  KVS_ADD_METADATA_VALUE, G_TYPE_STRING, value.c_str(), 
+                  KVS_ADD_METADATA_PERSISTENT, G_TYPE_BOOLEAN, persistent, NULL);
+  GstEvent* event = gst_event_new_custom(GST_EVENT_CUSTOM_DOWNSTREAM, metadata);
```
sirknightj, line 220: Can we add a log line here like "`emit fragment metadata event: <structure>`"?

niyatim23: Done!

hassanctech (approved):
LGTM

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -715,12 +792,16 @@ int gstreamer_init(int argc, char *argv[], CustomData *data) {
     g_signal_connect(G_OBJECT(bus), "message::error", (GCallback) error_cb, data);
     g_signal_connect(G_OBJECT(bus), "message::eos", G_CALLBACK(eos_cb), data);
     gst_object_unref(bus);
+
+    // Create a GStreamer timer to generate and put fragment metadata tags every 2 seconds
+    data->fragment_metadata_timer_id = g_timeout_add_seconds(data->put_fragment_metadata_frequency_seconds, reinterpret_cast<GSourceFunc>(put_metadata), kvssink);
```
sirknightj, line 797: Q: Does the fragment length needs to be above 2 seconds when using this sample?

src/gstreamer/gstkvssink.cpp
```diff
@@ -1161,29 +1162,30 @@ gst_kvs_sink_handle_sink_event (GstCollectPads *pads,
...
-            LOG_INFO("received kvs-add-metadata event for " << kvssink->stream_name);
-            if (NULL == gst_structure_get_string(structure, KVS_ADD_METADATA_NAME) ||
+            if (!gst_structure_has_name(structure, KVS_ADD_METADATA_G_STRUCT_NAME) || 
+                NULL == gst_structure_get_string(structure, KVS_ADD_METADATA_NAME) ||
                 NULL == gst_structure_get_string(structure, KVS_ADD_METADATA_VALUE) ||
                 !gst_structure_get_boolean(structure, KVS_ADD_METADATA_PERSISTENT, &persistent)) {
-
-                LOG_WARN("Event structure contains invalid field: " << std::string(gst_structure_to_string (structure)) << " for " << kvssink->stream_name);
+                ret = FALSE;
+                LOG_WARN("Event structure is invalid or it contains invalid field: " << std::string(gst_structure_to_string (structure)) << " for " << kvssink->stream_name);
```
sirknightj, line 1170: minor: "__an__ invalid field" or maybe "field(s)"

sirknightj (dismissed):

samples/kvssink_gstreamer_sample.cpp
niyatim23: Not necessarily because the first trigger would be as soon as this timer is added. This is modifiable by the customer as well

src/gstreamer/gstkvssink.cpp
niyatim23: Fixed!

hassanctech (approved):
lgtm

PR #1125, opened 2024-01-16
Ensure region for kvssink is set either through env or property

disa6302:
*Issue #, if available:*

*Description of changes:*
Before this change, the SDK set the default region to us-west-2. Because of this, if the env is set and the property is set, we would pick the env value, which is not correct. We should not override the property value set in the pipeline and check for env only if the property is not exclusively set to a value. To ensure this, the PR changes the default value to an empty string and enforces a region setting instead of defaulting to us-west-2. 

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -318,9 +318,9 @@ void kinesis_video_producer_init(GstKvsSink *kvssink)
         session_token_str = string(kvssink->session_token);
     }
 
-    if(IS_EMPTY_STRING(kvssink->aws_region)) {
+    if (IS_EMPTY_STRING(kvssink->aws_region)) {
         if (nullptr == (default_region = getenv(DEFAULT_REGION_ENV_VAR))) {
-            LOG_AND_THROW("No region set. Either set with env AWS_DEFAULT_REGION or set kvssink property aws-region");
+            LOG_AND_THROW("No region set. Either set with env " << DEFAULT_REGION_ENV_VAR << " or set kvssink property aws-region");
         } else {
             region_str = string(default_region);
         }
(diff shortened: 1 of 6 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -318,12 +318,15 @@ void kinesis_video_producer_init(GstKvsSink *kvssink)
         session_token_str = string(kvssink->session_token);
     }
 
-    if (nullptr == (default_region = getenv(DEFAULT_REGION_ENV_VAR))) {
-        region_str = string(kvssink->aws_region);
+    if(IS_EMPTY_STRING(kvssink->aws_region)) {
```
sirknightj, line 321:
```suggestion
    if (IS_EMPTY_STRING(kvssink->aws_region)) {
```

Also, let's use the GETENV macro from PIC.

For LOG_AND_THROW, can we have the string dynamically created based on the the env var and property names, and also include the fallback order if both are set in the message?

disa6302: Fixed. For property name, there doesnt seem to be a direct way to get the name other than running a for loop and comparing properties which seems wasteful. So leaving it as is.

sirknightj: I'm thinking we can extract `aws-region` string out as a macro

https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/de4299f942bb89c867d0dc73aeb4bc875d560a8f/src/gstreamer/gstkvssink.cpp#L612

sirknightj: Let's also update the description "AWS Region" to include the env var (other option).

disa6302: I think AWS region is ok. Adding env var is not relevant in the description of this property. Using a macro is fine, but I would like it to be consistent for all properties, which is beyond the scope of this PR.

sirknightj (approved):

PR #1126, opened 2024-01-16
Default to Release build

disa6302:
*Issue #, if available:*

*Description of changes:*
- The fix ensures the SDK is built in release mode by default. 
- Also updated log4cplus to 2.0.6 from 2.0.1 for non-windows platform since the library build fails on M1 in `CATCH` commands. This issue was fixed in 2.0.6: https://github.com/log4cplus/log4cplus/releases/tag/REL_2_0_6. For windows, the unit tests complete execution, but hangs while exiting and seems to be tied to upgrading log4cplus version. So keeping it at 2.0.1
- Removed jsmn cmake files since jsmn is included in producer C SDK and is unused in this SDK.

Testing:
- CI
- Confirmed the pipelines and the samples run fine


hassanctech (approved):
lgtm

PR #1131, opened 2024-01-23
CollectPadsStop to before parent state transition

stefankiesz:
The stopping of collect pads is an exception to the "downward state transitions after parent state transition" rule.

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1588,13 +1588,23 @@ gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
     /*
         The below state transition cases are separated into two switch blocks:
         one for upward (NULL->READY->PAUSED->PLAYING) transitions and one for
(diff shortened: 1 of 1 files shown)


stefankiesz:
No we don't have any kvssink tests at all, have just tested for functionality locally, and this change isn't addressing any actual issue anyone was seeing, rather it is just more closely adhering to GStreamer recommendations/guidelines.

hassanctech:
Can we at least run whatever `kvssink` samples we have to get some confidence that we are not regressing, whatever manual testing we have at our disposable with various pipelines that actually would exercise this code path.

stefankiesz:
Sounds good, the local testing so far has been just the kvssink sample on Mac, I'll now run all the [example command line pipelines](https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/examples-gstreamer-plugin.html#:~:text=plugin%20%27kvssink%27-,Example%20GStreamer%20launch%20commands,-The%20following%20examples) on Ubuntu, Windows, and Mac.

hassanctech (approved):
lgtm

disa6302 (approved):

PR #1151, opened 2024-03-05
remove unnecessary things from build

hassanctech:
*Issue #, if available:*

*Description of changes:*
- Remove unnecessary dependencies, this SDK does not have a direct dependency on curl, openssl
- Switch to cmake for log4cplus build instead of automake/autoconf, removed automake/autoconf from dependencies
- Remove included CPP Producer header file in the kvssink sample, it was needed only because a simple get time function was being called, instead call the C++ chrono library function directly.  This sample is meant to demonstrate usage of kvssink GStreamer element.  Client applications using `kvssink` are *not* expected to also have a dependency on the C++ Producer library since the `kvssink` plugin is meant to be loaded at runtime there are no header files from our SDK that need to be included in order to use `kvssink` GStreamer element.
- Fixes ci issue for mac os clang due to setting LDFLAGS which would choose system curl instead of the one we built
- Bring Windows Log4cplus version up to date with Linux, this required disabling threadpool for Windows since otherwise log4cplus internal threadpool fails to teardown properly
- Explicitly link samples to log4cplus since it includes header files from the logger


disa6302 (approved):
Awesome!

stefankiesz (approved):
Nice!

PR #1153, opened 2024-03-12
kvssink soft start/stop support

stefankiesz:
*Issue #, if available:*

### *What was changed?*
- Added a sample for intermittent streaming with kvssink that involves no tearing down of the gstreamer pipeline nor kvssink resources - a "soft" stop.
  - The sample supports device (`autovideosrc`) and test (`videotestsrc`) GStreamer sources
- Modified kvssink's handling and the behavior of EOS events
    - kvssink no longer sends an EOS message and event nor frees the KVS stream upon null buffers
    - kvssink does now send an EOS message to the bus upon receiving an EOS event
    - kvssink now sends an EOFR frame upon receiving an EOS message

### *Why was it changed?*
- Not sending an EOS upon receiving a null buffer more closely aligns kvssink with expected sink element behavior. It is typical for the source element of a pipeline to send EOS events down stream. The `x264enc` element will send null buffers to itself to clear its queue of buffered frames upon receiving an EOS event, so freeing KVS resources upon a null buffer would not allow for soft stops in this fashion.
- Upon receiving and handling an EOS event, sink elements should notify the bus via an appropriate message that the pipeline has completed the EOS. This proper handling of the EOS event allows the intermittent sample application to listen in for the EOS message to confirm when it will be safe resume the stream - we want the EOS event to have fully propagated down the pipeline before resuming the stream. The sending of an EOFR prevents the latest fragment prior to pausing from being held in the KVS buffer for the KVS intermittent wait duration.

### *How was it changed?*
The following is the flow of the new intermittent sample:
1. Play the pipeline.
1. Start a new thread to handle the start/stop mechanism at a specified interval.
2. Wait for the playing interval to finish using a conditional variable to allow for early exit with sigint.
3. Stop the stream by sending an EOS message on the pipeline.
4. Wait for a return EOS message to come on the bus once kvssink receives the EOS (this is on the separate, bus/main_loop thread).
5. Send a flush start event on the pipeline. This removes the EOS from sticking on the elements, but they still remain in a stopped state.
6. Wait for the paused interval to finish using a conditional variable to allow for early exit with sigint.
7. Send a flush stop event on the pipeline to resume streaming.

### *What testing was done for the changes?*
Testing for data loss was conducted by streaming footage of a running stop watch at a start/stop period of less than a fragment (6, 8, and 10 second intervals were tested). The time segment captured by the ingested KVS stream was compared to the desired start/stop increments. All fragments were within 100ms of the expected duration, except for the first fragments of streams which were ~300ms shorter than expected, likely due to time taken constructing resources.

Screenshot

<br>
<br>

Testing was also conducted with longer increments of 2min playing and 5min paused. The stream continued to stop and start with no issues. Below shows 1 hour of such a test with the orange-colored timeline showing where there were ingested fragments.

Screenshot

<br>
<br>


CMakeLists.txt
```diff
@@ -257,6 +257,9 @@ if(BUILD_GSTREAMER_PLUGIN)
   add_executable(kvs_gstreamer_file_uploader_sample samples/kvs_gstreamer_file_uploader_sample.cpp)
   target_link_libraries(kvs_gstreamer_file_uploader_sample ${LOG4CPLUS_LIBRARIES} ${GST_APP_LIBRARIES})
 
+  add_executable(kvssink_intermittent_sample samples/kvssink_intermittent_sample.cpp)
+  target_link_libraries(kvssink_intermittent_sample ${LOG4CPLUS_LIBRARIES} ${GST_APP_LIBRARIES} kvspic)
```
disa6302, line 261: Why link kvspic?

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+    StreamSource source_type;
+    char stream_name[MAX_STREAM_NAME_LEN + 1];
+
+    gst_init (&argc, &argv);
+
+
+    /* Parse input arguments */
+
+    // Check for invalid argument count.
+    if (argc > 3 || argc < 2) {
```
disa6302, line 169: Would suggest splitting this into 2 checks to clarify what is missing.

disa6302 (commented):
Awesome!

samples/kvssink_intermittent_sample.cpp
sirknightj, line 169: More readable way:
```
argc != 2 && argc != 3
```

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+    /* Create GStreamer elements */
+
+    pipeline = gst_pipeline_new ("kvs-pipeline");
+
+    /* source */
+    if(source_type == TEST_SOURCE) {
+        source = gst_element_factory_make ("videotestsrc", "test-source");
+        g_object_set(G_OBJECT(source), "is-live", TRUE, NULL);
+    } else if (source_type == DEVICE_SOURCE) {
+        source = gst_element_factory_make ("autovideosrc", "device-source");
```
sirknightj, line 207: Some places have a space between method and parenthesis and others not - please fix

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+    /* source filter */
+    source_filter = gst_element_factory_make("capsfilter", "source-filter");
+    source_caps = gst_caps_new_simple("video/x-raw", "format", G_TYPE_STRING, "I420", NULL);
+    g_object_set(G_OBJECT(source_filter), "caps", source_caps, NULL);
+    gst_caps_unref(source_caps);
+
+    /* encoder */
+    encoder = gst_element_factory_make("x264enc", "encoder");
+    g_object_set(G_OBJECT(encoder), "bframes", 0, NULL);
```
sirknightj, line 224: We should also set key int max to make sure too long fragments don't get generated

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+    if (!pipeline || !source || !clock_overlay || !video_convert || !source_filter || !encoder || !sink_filter) {
+        LOG_ERROR("[KVS sample] Not all GStreamer elements could be created.");
+        return -1;
+    }
+
+    // Populate data struct.
+    customData.main_loop = main_loop;
+    customData.pipeline = pipeline;
+
+    //Add a message handler.
```
sirknightj, line 257: Space

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+    bus_watch_id = gst_bus_add_watch (bus, bus_call, &customData);
+    gst_object_unref (bus);
+
+    // Add elements into the pipeline.
+    gst_bin_add_many (GST_BIN (pipeline),
+                        source, clock_overlay, video_convert, source_filter, encoder, sink_filter, kvssink, NULL);
+
+    // Link the elements together.
+    if (!gst_element_link_many(source, clock_overlay, video_convert, source_filter, encoder, sink_filter, kvssink, NULL)) {
+            LOG_ERROR("[KVS sample] Elements could not be linked");
```
sirknightj, line 268: Indent

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+#include <glib.h>
+
+#include "gstreamer/gstkvssink.h"
+
+using namespace com::amazonaws::kinesis::video;
+using namespace log4cplus;
+
+/* modify these values to change start/stop interval */
+#define KVS_INTERMITTENT_PLAYING_INTERVAL_SECONDS 20
+#define KVS_INTERMITTENT_PAUSED_INTERVAL_SECONDS 40
```
sirknightj, line 16: Can we add usage of this sample to the readme + also explain high level what it does (streams to kvs for 20 seconds then stops streaming for 40 seconds, and repeats the cycle indefinitely)

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+}
+
+void determine_aws_credentials(GstElement *kvssink, char* streamName) {
+    char const *iot_credential_endpoint;
+    char const *cert_path;
+    char const *private_key_path;
+    char const *role_alias;
+    char const *ca_cert_path;
+    char const *credential_path;
+    if (nullptr != (iot_credential_endpoint = getenv("IOT_GET_CREDENTIAL_ENDPOINT")) &&
```
sirknightj, line 90: Use capital GETENV (macro) from PIC.

We also have IS_NULL_OR_EMPTY_STRING macro as well.

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+            "ca-path", G_TYPE_STRING, ca_cert_path,
+            "role-aliases", G_TYPE_STRING, role_alias, NULL);
+    
+    g_object_set(G_OBJECT (kvssink), "iot-certificate", iot_credentials, NULL);
+        gst_structure_free(iot_credentials);
+    // kvssink will search for long term credentials in envvar automatically so no need to include here
+    // if no long credentials or IoT credentials provided will look for credential file as last resort.
+    } else if(nullptr != (credential_path = getenv("AWS_CREDENTIAL_PATH"))){
+        g_object_set(G_OBJECT (kvssink), "credential-path", credential_path, NULL);
+    }
```
sirknightj, line 111: Can we add some debug logs in each branch to say which credentials path was chosen.

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,293 @@
...
+    /* source */
+    if(source_type == TEST_SOURCE) {
+        source = gst_element_factory_make ("videotestsrc", "test-source");
+        g_object_set(G_OBJECT(source), "is-live", TRUE, NULL);
+    } else if (source_type == DEVICE_SOURCE) {
+        source = gst_element_factory_make ("autovideosrc", "device-source");
+    } 
+
+    /* clock overlay */
+    clock_overlay = gst_element_factory_make("clockoverlay", "clock_overlay");
```
sirknightj, line 211: Can we add the date to clockoverlay? You can use the string from here: https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/gs-send-data.html

CMakeLists.txt
stefankiesz: I actually had to add `kvspicUtils` to this sample for it to build on Linux platforms.

samples/kvssink_intermittent_sample.cpp
stefankiesz: Actually, opted for allowing for no extra args to use kvssink default stream name.

README.md
```diff
@@ -190,7 +190,25 @@ The kvssink element has the following required parameters:
...
+
+#### kvssink Intermittent Sample
+Usage:
+```bash
+./kvssink_intermittent_sample <stream-name (optional)> <testsrc or devicesrc (optional)>
+```
+Setting the source to `testsrc` will use [videotestsrc](https://gstreamer.freedesktop.org/documentation/videotestsrc/?gi-language=c) and to `devicesrc` will use [autovideosrc](https://gstreamer.freedesktop.org/documentation/autodetect/autovideosrc.html?gi-language=c). By default, kvssink uses "DEFAULT_STREAM" as the stream name, and the sample uses videotestsrc as the source. If a stream with the provided or default name does not exist, the stream will automatically be created.
+
+The intermittent kvssink sample will stream video for 20 seconds, then pause for 40 seconds, and repeat until an interrupt signal is received. To manually adjust the streaming and paused intervals, you can change the `KVS_INTERMITTENT_PLAYING_INTERVAL_SECONDS` and `KVS_INTERMITTENT_PAUSED_INTERVAL_SECONDS` values in the *kvssink_intermittent_sample.cpp* file.
+
```
sirknightj, line 211: Nice!

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,323 @@
...
+    /* Parse input arguments */
+
+    // Check for invalid argument count, get stream name.
+    if(argc > 3) {
+        LOG_ERROR("[KVS sample] Invalid argument count, too many arguments.");
+        LOG_INFO("[KVS sample] Usage: " << argv[0] << " <streamName (optional)> <testsrc or devicesrc (optional)>");
+        return -1;
+    } else if(argc > 1) {
+        STRNCPY(stream_name, argv[1], MAX_STREAM_NAME_LEN);
+        stream_name[MAX_STREAM_NAME_LEN] = '\0';
```
sirknightj, line 195: Is this needed? You already set this on line 181.

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,323 @@
...
+    sink_caps = gst_caps_new_simple("video/x-h264",
+                                    "stream-format", G_TYPE_STRING, "avc",
+                                    "alignment", G_TYPE_STRING, "au",
+                                    NULL);
+    g_object_set(G_OBJECT(sink_filter), "caps", sink_caps, NULL);
+    gst_caps_unref(sink_caps);
+
+    /* kvssink */
+    kvssink = gst_element_factory_make("kvssink", "kvssink");
+    if (stream_name[0] != '\0') {
```
sirknightj, line 263: !IS_EMPTY_STRING()

just in case: you can test this by passing in `./sampleName ""`

src/gstreamer/gstkvssink.cpp
```diff
@@ -1188,6 +1188,23 @@ gst_kvs_sink_handle_sink_event (GstCollectPads *pads,
         }
         case GST_EVENT_EOS: {
             LOG_INFO("EOS Event received in sink for " << kvssink->stream_name);
+            
+            if(data && data->kinesis_video_stream) {
+                Frame eofr = EOFR_FRAME_INITIALIZER;
+                STATUS put_eofr_status = data->kinesis_video_stream->putFrame(eofr);
```
sirknightj, line 1194: Can we add a comment saying that end of fragment was added.

src/gstreamer/gstkvssink.cpp
```diff
@@ -1188,6 +1188,23 @@ gst_kvs_sink_handle_sink_event (GstCollectPads *pads,
...
+                if(STATUS_FAILED(put_eofr_status)) {
+                    LOG_WARN("Failed to put EOFR for " << kvssink->stream_name);
+                }
+            } else {
+                LOG_WARN("NULL pointer! Failed to put EOFR for " << kvssink->stream_name);
```
sirknightj, line 1199: Saying null ptr in warn looks scary & looks like an error. This is handling the case where kinesis_video_stream failed to initialize?

samples/kvssink_intermittent_sample.cpp
stefankiesz: Correct, not needed now that the whole buffer is being initialized as null. Will remove

src/gstreamer/gstkvssink.cpp
stefankiesz: Yes and as a safety measure in case either of these two pointers get nulled. This correlates with a STATUS_NULL_ARG error we get from Producer for these kind of null checks. How does "Null argument, failed to put EOFR for " sound?

sirknightj: I'm thinking that we should just make this an INFO log and say what it's doing literally.

"Kinesis Video Stream failed to initialize, not putting EOFR".

samples/kvssink_intermittent_sample.cpp
```diff
@@ -0,0 +1,322 @@
...
+    _CustomData() :
+            main_loop(NULL),
+            pipeline(NULL) {}
+
+    GMainLoop *main_loop;
+    GstElement *pipeline;
+} CustomData;
+
+void sigint_handler(int sigint) {
+    LOG_DEBUG("SIGINT received.  Exiting...");
```
sirknightj, line 43: /nit 2 spaces

sirknightj (approved):

src/gstreamer/gstkvssink.cpp
stefankiesz: I'm still leaning towards having it be a WARN since it 'failed' to initialize. @disa6302 any thoughts?

sirknightj (approved):

src/gstreamer/gstkvssink.cpp
```diff
@@ -1664,15 +1661,8 @@ gst_kvs_sink_change_state(GstElement *element, GstStateChange transition) {
...
-            // Need this check in case an EOS was received in the buffer handler and
-            // stream was already stopped. Although stopSync() is an idempotent call,
-            // we want to avoid an extra call
-            if(!data->streamingStopped.load()) {
-                data->kinesis_video_stream->stopSync();
-                data->streamingStopped.store(true);
-            } else {
-                LOG_INFO("Streaming already stopped for " << kvssink->stream_name);
-            }
+            data->kinesis_video_stream->stopSync();
```
unicornss, line 1664: suggest adding a log to denote trying to stop.

unicornss (approved):

stefankiesz:
This PR is based on an older develop branch. Closing in favor of https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1257

src/gstreamer/gstkvssink.cpp
stefankiesz: This log is already there in the above switch block:

```
// (Downward Transition) gst_collect_pads_stop must be called prior to parent class PAUSED->READY transition.
case GST_STATE_CHANGE_PAUSED_TO_READY:
    LOG_INFO("Stopping kvssink for " << kvssink->stream_name);
```

PR #1157, opened 2024-03-21
Bump GStreamer Version in Windows CI

stefankiesz:
*Issue #, if available:*

*What was changed?*
- Updated GStreamer version installed on Windows CI runner from 1.16.2 to the latest 1.22.8
- Updated pkg-config.exe path as there is a folder name change in between the GStreamer versions

*Why was it changed?*
- The 1.16.2 version is now missing from their site: https://gstreamer.freedesktop.org/data/pkg/windows/
- 1.16.3 is present, and the base package successfully downloaded, but not the gstreamer-devel package. The latest version 1.22.8, however, is working.

*How was it changed?*
- Changed the version number is the CI install step
- Modified the path of where to find pkg-config for Windows in the CMakeLists

- The CI was run to verify all are passing

<br>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,8 +322,8 @@ jobs:
       - name: Install dependencies
         run: |
           choco install nasm strawberryperl
-          choco install gstreamer --version=1.16.3
-          choco install gstreamer-devel --version=1.16.3
+          choco install gstreamer
+          choco install gstreamer-devel
       - name: Build repository
         run: |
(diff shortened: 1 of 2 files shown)


sirknightj:
image
https://gstreamer.freedesktop.org/pkg/windows/1.16.2/ it definitely did exist before. 

1.16 is past EOL though so that is probably why it was removed https://endoflife.date/gstreamer

sirknightj (approved):

disa6302 (approved):

PR #1160, opened 2024-03-28
Add image tag java

stefankiesz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
+++ b/src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
@@ -485,10 +485,10 @@ void KinesisVideoClientWrapper::putKinesisVideoEventMetadata(jlong streamHandle,
         return;
     }
 
-    if (pStreamEventMetadata == NULL)
+    if (kinesisStreamEventMetadata == NULL)
     {
-        DLOGE("pStreamEventMetadata is NULL");
-        throwNativeException(env, EXCEPTION_NAME, "pStreamEventMetadata is NULL.", STATUS_INVALID_OPERATION);
+        DLOGE("kinesisStreamEventMetadata is NULL");
+        throwNativeException(env, EXCEPTION_NAME, "kinesisStreamEventMetadata is NULL.", STATUS_INVALID_OPERATION);
         return;
     }
 
--- a/src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
+++ b/src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
@@ -453,11 +453,62 @@ void KinesisVideoClientWrapper::putKinesisVideoFragmentMetadata(jlong streamHand
 
     if (STATUS_FAILED(retStatus))
     {
-        DLOGE("Failed to put a metadata with status code 0x%08x", retStatus);
-        throwNativeException(env, EXCEPTION_NAME, "Failed to put a metadata into the stream.", retStatus);
+        DLOGE("Failed to put a fragment metadata with status code 0x%08x", retStatus);
+        throwNativeException(env, EXCEPTION_NAME, "Failed to put a fragment metadata into the stream.", retStatus);
+        return;
+    }
+}
+
+void KinesisVideoClientWrapper::putKinesisVideoEventMetadata(jlong streamHandle, jint event, jobject kinesisStreamEventMetadata)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    JNIEnv *env;
+    mJvm->GetEnv((PVOID*) &env, JNI_VERSION_1_6);
+
+    if (!IS_VALID_CLIENT_HANDLE(mClientHandle))
+    {
+        DLOGE("Invalid client object");
+        throwNativeException(env, EXCEPTION_NAME, "Invalid call after the client is freed.", STATUS_INVALID_OPERATION);
         return;
     }
 
+    if (!IS_VALID_STREAM_HANDLE(streamHandle))
+    {
+        DLOGE("Invalid stream handle 0x%016" PRIx64, (UINT64) streamHandle);
+        throwNativeException(env, EXCEPTION_NAME, "Invalid stream handle.", STATUS_INVALID_OPERATION);
+        return;
+    }
+
+    if (event >= STREAM_EVENT_TYPE_LAST || event == STREAM_EVENT_TYPE_NONE) {
+        DLOGE("Stream event is invalid");
+        throwNativeException(env, EXCEPTION_NAME, "Stream event is invalid.", STATUS_INVALID_OPERATION);
+        return;
+    }
+
+    if (pStreamEventMetadata == NULL)
+    {
+        DLOGE("pStreamEventMetadata is NULL");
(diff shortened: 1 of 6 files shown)


src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
```diff
@@ -456,11 +456,86 @@ void KinesisVideoClientWrapper::putKinesisVideoFragmentMetadata(jlong streamHand
...
 
+    if (!IS_VALID_STREAM_HANDLE(streamHandle))
+    {
+        DLOGE("Invalid stream handle 0x%016" PRIx64, (UINT64) streamHandle);
+        throwNativeException(env, EXCEPTION_NAME, "Invalid stream handle.", STATUS_INVALID_OPERATION);
+        return;
+    }
+
+    // Check for a negative STREAM_EVENT_TYPE coming from Java.
+    if (event < STREAM_EVENT_TYPE_NONE) {
```
niyatim23, line 486: Can we also check the upper limit here?

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -1188,3 +1193,131 @@ BOOL releaseStreamDataBuffer(JNIEnv* env, jobject dataBuffer, UINT32 offset, PBY
...
+    methodId = env->GetMethodID(cls, "getImagePrefix", "()Ljava/lang/String;");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getImagePrefix");
+    } else {
+        jstring retString = (jstring) env->CallObjectMethod(streamEventMetadata, methodId);
+        CHK_JVM_EXCEPTION(env);
+
+        if (retString != NULL) {
+            jsize javaStringLength = env->GetStringUTFLength(retString);
+            CHK(javaStringLength <= MAX_IMAGE_PREFIX_LENGTH, STATUS_INVALID_ARG_LEN);
```
niyatim23, line 1230: I think this check should be in the `putKinesisVideoEventMetadata` in Producer C SDK. Did you add it here because it was missing there?

src/JNI/include/com/amazonaws/kinesis/video/producer/jni/Parameters.h
```diff
@@ -16,6 +16,8 @@ BOOL setStreamingEndpoint(JNIEnv* env, jstring streamingEndpoint, PCHAR pEndpoin
 BOOL setStreamDataBuffer(JNIEnv* env, jobject dataBuffer, UINT32 offset, PBYTE* ppBuffer);
 BOOL releaseStreamDataBuffer(JNIEnv* env, jobject dataBuffer, UINT32 offset, PBYTE pBuffer);
 BOOL setTags(JNIEnv *env, jobjectArray tagArray, PTag* ppTags, PUINT32 pTagCount);
+BOOL setStreamEventMetadata(JNIEnv* env, jobject streamEventMetadata, PStreamEventMetadata pStreamEventMetadata);
+BOOL allocStreamEventMetadataArray(JNIEnv* env, jobject streamEventMetadata, PCHAR* metaDataArray, jmethodID methodId);
```
niyatim23, line 20: Do we need to declare `allocStreamEventMetadataArray` here if it is a helper function for `setStreamEventMetadata` and not used anywhere else? Can we just define it directly in Parameters.cpp?

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -1188,3 +1193,131 @@ BOOL releaseStreamDataBuffer(JNIEnv* env, jobject dataBuffer, UINT32 offset, PBY
...
+    CHK(namesArrayLength <= MAX_EVENT_CUSTOM_PAIRS, STATUS_MAX_FRAGMENT_METADATA_COUNT);
+
+    // Null all elements for safety.
+    MEMSET(metaDataArray, NULL, sizeof(PCHAR) * MAX_EVENT_CUSTOM_PAIRS);
+            
+    // Iterate through the char pointers, allocating memory for the Java string that will be copied to the char pointers.
+    for (jsize i = 0; i < namesArrayLength; i++) {
+        stringElement = (jstring) env->GetObjectArrayElement(retArray, i);
+        jsize javaStringLength = env->GetStringUTFLength(stringElement);
+        CHK(javaStringLength >= 0, STATUS_INVALID_ARG); // (jsize is signed, but will be used as an unsigned SIZE_T)
```
niyatim23, line 1293: The `putKinesisVideoEventMetadata` API should allow values to empty or have length=0 because that is one of the ways to unset / remove persistent metadata. I'm not sure about name being empty but you may want to check that as well

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -1188,3 +1193,131 @@ BOOL releaseStreamDataBuffer(JNIEnv* env, jobject dataBuffer, UINT32 offset, PBY
...
+    }
+
+CleanUp:
+    return STATUS_FAILED(retStatus) ? FALSE : TRUE;
+}
+
+BOOL allocStreamEventMetadataArray(JNIEnv* env, jobject streamEventMetadata, PCHAR* metaDataArray, jmethodID methodId) {
+    STATUS retStatus = STATUS_SUCCESS;
+    jobjectArray retArray = NULL;
+    jsize namesArrayLength = 0;
```
niyatim23, line 1271: Recommend renaming this to something generic as this function is used for name and value both

src/JNI/com/amazonaws/kinesis/video/producer/jni/KinesisVideoClientWrapper.cpp
stefankiesz: No, we want to keep the upper limit check in PIC/ProducerC as we do for most of the existing parameters here in the JNI. However, we have to check for a negative value here as it is a signed int coming from Java.

src/JNI/include/com/amazonaws/kinesis/video/producer/jni/Parameters.h
stefankiesz: Yes, but then the order of the functions are placed in the file will matter, I'm ok with either way.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
stefankiesz: Yes nice catch!

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
stefankiesz: No, it is there, I overlooked removing this when I was cleaning up the extra checks I had here, nice catch.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
stefankiesz: Btw, this _event_ metadata doesn't have the functionality of persistency, but it looks like I have already set it to allow a 0 value, maybe I did this after you added this comment. Can you please verify that `javaStringLength >= 0` is good to go and is what you were referring to, thanks.

PR #1161, opened 2024-04-02
Fix un-initialized JNI ClientInfo members 

stefankiesz:
### *What was changed?*
The previously un-initialized `pClientInfo` members in the JNI are now initialized to NULL values.

### *Why was it changed?*
The `kvsRetryStrategyCallbacks` function pointers not being initialized as NULL was causing the function pointers to not be assigned in the PIC layer which was then causing a segmentation fault in the Producer C layer when running the Java Producer SDK.

Here in PIC, at least one of the `pKvsRetryStrategyCallbacks` function pointers must be NULL in order for `setupDefaultKvsRetryStrategyParameters` to be called to setup the default callbacks, 

```
if (pKvsRetryStrategyCallbacks->createRetryStrategyFn == NULL || pKvsRetryStrategyCallbacks->freeRetryStrategyFn == NULL ||
pKvsRetryStrategyCallbacks->executeRetryStrategyFn == NULL || pKvsRetryStrategyCallbacks->getCurrentRetryAttemptNumberFn == NULL) {
        CHK_STATUS(setupDefaultKvsRetryStrategyParameters(pKinesisVideoClient));
}
```

but the function pointers were never explicitly initialized as NULL in the JNI, so they would usually all be junk, non-null values,

```
[testing] createRetryStrategyFn: 0xb33300b20100b3
[testing] freeRetryStrategyFn: 0x180062000000
[testing] executeRetryStrategyFn: 0x22001c0011001b00
[testing] getCurrentRetryAttemptNumberFn: 0x960001000000b174
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0xa) at pc=0x00b33300b20100b3, pid=17942, tid=10243
```
causing the seg-fault to occur when the undefined `createRetryStrategyFn` is invoked in the below Producer C code,
```
CHK_STATUS(pKvsRetryStrategyCallbacks->createRetryStrategyFn(&(pKinesisVideoClient->deviceInfo.clientInfo.kvsRetryStrategy)));
```
and when the app did work, all the pointers happened to be NULL:

```
[testing] createRetryStrategyFn: 0x0
[testing] freeRetryStrategyFn: 0x0
[testing] executeRetryStrategyFn: 0x0
[testing] getCurrentRetryAttemptNumberFn: 0x0
[testing] Calling setupDefaultKvsRetryStrategyParameters
```

### *How was it changed?*
Carefully, and by initializing previously un-initialized `pClientInfo` members in the JNI to NULL values.

### *What testing was done for the changes?*
The Java sample successfully ran ~10 times using the outputted Producer CPP JNI with no issues. Previously, the application would seg-fault every ~5/6 runs.

<br> 
<br> 

### DIFF
--- a/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
+++ b/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
@@ -280,15 +280,6 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
         }
     }
     
-    // Note: kvsRetryStrategyCallbacks is not supported yet in Java, the below is a skeleton for the future implementation.
-    methodId = env->GetMethodID(cls, "getKvsRetryStrategyCallbacks", "()Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategyCallbacks;");
-    if (methodId == NULL) {
-        DLOGW("Couldn't find method id getKvsRetryStrategyCallbacks");
-    } else {
-        kvsRetryStrategyCallbacks = (jobject) env->CallObjectMethod(clientInfo, methodId);
-        CHK_JVM_EXCEPTION(env);
-    }
-
 CleanUp:
     return STATUS_FAILED(retStatus) ? FALSE : TRUE;
 }
@@ -299,18 +290,11 @@ BOOL setKvsRetryStrategy(JNIEnv *env, jobject kvsRetryStrategy, PKvsRetryStrateg
     jmethodID methodId = NULL;
     jclass cls = NULL;
 
-    CHECK(env != NULL && pKvsRetryStrategy != NULL);
-
-    if (kvsRetryStrategy == NULL) {
-        goto CleanUp;
-    }
+    CHK(env != NULL && pKvsRetryStrategy != NULL, STATUS_NULL_ARG);
+    CHK_WARN(kvsRetryStrategy != NULL, STATUS_INVALID_OPERATION, "Failed to get Java kvsRetryStrategy class.");
 
     cls = env->GetObjectClass(kvsRetryStrategy);
-    if (cls == NULL) {
-        DLOGW("Failed to create Java kvsRetryStrategy class.");
-        CHK(FALSE, STATUS_INVALID_OPERATION);
-        goto CleanUp;
-    }
+    CHK_WARN(cls != NULL, STATUS_INVALID_OPERATION, "Failed to create Java kvsRetryStrategy class.");
 
     methodId = env->GetMethodID(cls, "getRetryStrategy", "()J");
     if (methodId == NULL) {
@@ -328,10 +312,10 @@ BOOL setKvsRetryStrategy(JNIEnv *env, jobject kvsRetryStrategy, PKvsRetryStrateg
         CHK_JVM_EXCEPTION(env);
     }
 
+    // TODO: Test with logs that this is getting the value from Java, play around with the value.
     methodId = env->GetMethodID(cls, "getRetryStrategyType", "()I");
     if (methodId == NULL) {
         DLOGW("Couldn't find method id getRetryStrategyType, setting retryStrategyType to EXPONENTIAL_BACKOFF_WAIT.");
-        pKvsRetryStrategy->retryStrategyType = KVS_RETRY_STRATEGY_EXPONENTIAL_BACKOFF_WAIT;
     } else {
         pKvsRetryStrategy->retryStrategyType = (KVS_RETRY_STRATEGY_TYPE) env->CallIntMethod(kvsRetryStrategy, methodId);
         CHK_JVM_EXCEPTION(env);
--- a/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
+++ b/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
@@ -158,6 +158,7 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
     STATUS retStatus = STATUS_SUCCESS;
     jmethodID methodId = NULL;
     const char *retChars;
+    jobject kvsRetryStrategyCallbacks = NULL;
 
     CHECK(env != NULL && clientInfo != NULL && pClientInfo != NULL);
 
@@ -256,7 +257,6 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
         CHK_JVM_EXCEPTION(env);
     }
     
-    jobject kvsRetryStrategyCallbacks = NULL;
     methodId = env->GetMethodID(cls, "getKvsRetryStrategyCallbacks", "()Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategyCallbacks;");
     if (methodId == NULL) {
         DLOGW("Couldn't find method id getKvsRetryStrategyCallbacks");
@@ -281,16 +281,16 @@ VOID setKvsRetryStrategyCallbacks(JNIEnv *env, jobject kvsRetryStrategyCallbacks
         goto CleanUp;
     }
 
-    jclass cls = env->GetObjectClass(kvsRetryStrategyCallbacks);
+    cls = env->GetObjectClass(kvsRetryStrategyCallbacks);
     if (cls == NULL) {
         DLOGW("Failed to create Java kvsRetryStrategyCallbacks class.");
         nullSetCallbacks = TRUE;
         goto CleanUp;
     }
 
-.   /* *** */
+    /* *** */
     /* NOTE: All kvsRetryStrategyCallbacks must be set to NULL in order for PIC to set them to the default callbacks.*/
-.   /* *** */
+    /* *** */
 
     methodId = env->GetMethodID(cls, "createRetryStrategyFn", "(Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategyCallbacks;)V");
     if (methodId == NULL) {
--- a/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
+++ b/src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
@@ -146,6 +146,7 @@ BOOL setDeviceInfo(JNIEnv *env, jobject deviceInfo, PDeviceInfo pDeviceInfo)
         CHK_JVM_EXCEPTION(env);
 
         if (!setClientInfo(env, clientInfo, &pDeviceInfo->clientInfo)) {
+            // TODO: Consider hard failing here rather than just logging.
             DLOGW("Failed getting/setting client info.");
         }
     }
@@ -254,7 +255,6 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
     methodId = env->GetMethodID(cls, "getMetricLoggingPeriod", "()J");
     if (methodId == NULL) {
         DLOGW("Couldn't find method id getMetricLoggingPeriod");
-        pClientInfo->metricLoggingPeriod = 0;
     } else {
         pClientInfo->metricLoggingPeriod = env->CallLongMethod(clientInfo, methodId);
         CHK_JVM_EXCEPTION(env);
@@ -263,7 +263,6 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
     methodId = env->GetMethodID(cls, "getReservedCallbackPeriod", "()J");
     if (methodId == NULL) {
         DLOGW("Couldn't find method id getReservedCallbackPeriod");
-        pClientInfo->reservedCallbackPeriod = 0;
     } else {
         pClientInfo->reservedCallbackPeriod = env->CallLongMethod(clientInfo, methodId);
         CHK_JVM_EXCEPTION(env);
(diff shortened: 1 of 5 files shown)


stefankiesz:
Note to self: Testing shows JNI is setting the retry callbacks to null at this point, confirmed by seeing all four `Couldn't find method id <type>RetryStrategyFn, setting to NULL` logs

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -161,6 +161,17 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
 
     CHECK(env != NULL && clientInfo != NULL && pClientInfo != NULL);
 
+    // Initialize Java-unused members. These will be set to default values in PIC.
+    pClientInfo->kvsRetryStrategy.pRetryStrategy = NULL;
+    pClientInfo->kvsRetryStrategy.pRetryStrategyConfig = NULL;
+    pClientInfo->kvsRetryStrategy.retryStrategyType = KVS_RETRY_STRATEGY_DISABLED;
```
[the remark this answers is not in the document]
stefankiesz: Updated with getters per our offline discussion

niyatim23 (commented):
Refer [this](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1965/files#diff-9be07c31f5e625a3878fb4d65f2481b851fcfdf185d09813ef9222c63a8bc44aR7) to fix the Windows build

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,176 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+    methodId = env->GetMethodID(cls, "getReservedCallbackPeriod", "()J");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getReservedCallbackPeriod");
+        pClientInfo->reservedCallbackPeriod = 0;
+    } else {
+        pClientInfo->reservedCallbackPeriod = env->CallLongMethod(clientInfo, methodId);
+        CHK_JVM_EXCEPTION(env);
+    }
+
+    methodId = env->GetMethodID(cls, "getKvsRetryStrategy", "()Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategy;");
```
niyatim23, line 272: Couldn't find this function in the Java layer

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,176 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
         CHK_JVM_EXCEPTION(env);
     }
 
+    methodId = env->GetMethodID(cls, "getMetricLoggingPeriod", "()J");
```
niyatim23, line 254: I couldn't find this method in the Java SDK. Is there something that I'm missing here?

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,176 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getMetricLoggingPeriod");
+        pClientInfo->metricLoggingPeriod = 0;
+    } else {
+        pClientInfo->metricLoggingPeriod = env->CallLongMethod(clientInfo, methodId);
+        CHK_JVM_EXCEPTION(env);
+    }
+
+    methodId = env->GetMethodID(cls, "getReservedCallbackPeriod", "()J");
```
niyatim23, line 263: I couldn't find this method in the Java SDK. Is there something that I'm missing here?

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
disa6302: +1. We need this and getReservedCallbackPeriod in your Java PR.

stefankiesz: These changes were intended to work to null-initialize the currently not initialized members without any changes being made in the Java end. This PR was not to add support for theses members and classes in Java, but to prevent undefined behavior from these members not being null-initialized. Will discuss more offline in case we do want this to support setting/getting members from Java.

niyatim23: I understand, I'm not sure if it is a good idea to look for functions that don't exist in the Java layer. Can we initialize it without looking for that function instead? Because it is easy to misunderstand this as something missing

stefankiesz: That's what my changes were doing before we decided to add in all this framework to be able to readily hook it up to the Java end, and the Java-side changes were just for parity purposes just to be used as reference. It looks like we do want to somewhat implement this in the Java end now though, so I'm working on setting up Java getters that simply return null rather than the expected class object.

niyatim23: Understood, thanks for taking care of this!

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
niyatim23: Understood, thanks for taking care of this!

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,176 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+
+    cls = env->GetObjectClass(kvsRetryStrategy);
+    if (cls == NULL) {
+        DLOGW("Failed to create Java kvsRetryStrategy class.");
+        nullKvsRetryStrategy = TRUE;
+        goto CleanUp;
+    }
+
+    methodId = env->GetMethodID(cls, "getRetryStrategy", "()Lcom/amazonaws/kinesisvideo/producer/RetryStrategy");
+    if (methodId == NULL) {
```
stefankiesz, line 315: Add comments stating these return null from Java - not the member

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+        kvsRetryStrategyCallbacks = (jobject) env->CallObjectMethod(clientInfo, methodId);
+        CHK_JVM_EXCEPTION(env);
+    }
+    setKvsRetryStrategyCallbacks(env, kvsRetryStrategyCallbacks, &pClientInfo->kvsRetryStrategyCallbacks);
+
 CleanUp:
     return STATUS_FAILED(retStatus) ? FALSE : TRUE;
 }
 
+VOID setKvsRetryStrategy(JNIEnv *env, jobject kvsRetryStrategy, PKvsRetryStrategy pKvsRetryStrategy)
```
stefankiesz, line 294: This successfully set to the desired default KVS_RETRY_STRATEGY_EXPONENTIAL_BACKOFF_WAIT as tested by turning off the wifi mid-stream and seeing these logs specify the retry type as (1):
```
KinesisVideoStream base result is [400]. Executing KVS retry handler of retry strategy type [1]
```

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+    }
+
+    methodId = env->GetMethodID(cls, "getKvsRetryStrategy", "()Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategy;");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getKvsRetryStrategy");
+    } else {
+        kvsRetryStrategy = (jobject) env->CallObjectMethod(clientInfo, methodId);
+        CHK_JVM_EXCEPTION(env);
+    }
+    setKvsRetryStrategy(env, kvsRetryStrategy, &pClientInfo->kvsRetryStrategy);
```
disa6302, line 279: This should be after line 277. We should do this only if the method is found as discussed offline.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+    
+    methodId = env->GetMethodID(cls, "getKvsRetryStrategyCallbacks", "()Lcom/amazonaws/kinesisvideo/producer/KvsRetryStrategyCallbacks;");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getKvsRetryStrategyCallbacks");
+    } else {
+        kvsRetryStrategyCallbacks = (jobject) env->CallObjectMethod(clientInfo, methodId);
+        CHK_JVM_EXCEPTION(env);
+    }
+    setKvsRetryStrategyCallbacks(env, kvsRetryStrategyCallbacks, &pClientInfo->kvsRetryStrategyCallbacks);
```
disa6302, line 288: Same here.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
disa6302: Awesome!

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+        nullKvsRetryStrategy = TRUE;
+        goto CleanUp;
+    }
+
+    methodId = env->GetMethodID(cls, "getRetryStrategy", "()Lcom/amazonaws/kinesisvideo/producer/RetryStrategy");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getRetryStrategy, setting pRetryStrategy to null.");
+        pKvsRetryStrategy->pRetryStrategy = NULL;
+    } else {
+        if (env->CallLongMethod(kvsRetryStrategy, methodId) == 0) {
```
disa6302, line 319: Why isnt this following the same pattern as the rest. We need to check JVM exception and error out. Not silence the errors.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+            pKvsRetryStrategy->pRetryStrategy = NULL;
+        }
+    }
+    
+    methodId = env->GetMethodID(cls, "getRetryStrategyConfig", "()Lcom/amazonaws/kinesisvideo/producer/RetryStrategyConfig");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getRetryStrategyConfig, setting pRetryStrategyConfig to null.");
+        pKvsRetryStrategy->pRetryStrategyConfig = NULL;
+    } else {
+        if (env->CallLongMethod(kvsRetryStrategy, methodId) == 0) {
```
disa6302, line 332: Same here.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+            pKvsRetryStrategy->pRetryStrategyConfig = NULL;
+        }
+    }
+
+    methodId = env->GetMethodID(cls, "getRetryStrategyType", "()Lcom/amazonaws/kinesisvideo/producer/RetryStrategyType");
+    if (methodId == NULL) {
+        DLOGW("Couldn't find method id getRetryStrategyType, setting retryStrategyType to EXPONENTIAL_BACKOFF_WAIT.");
+        pKvsRetryStrategy->retryStrategyType = KVS_RETRY_STRATEGY_EXPONENTIAL_BACKOFF_WAIT;
+    } else {
+        pKvsRetryStrategy->retryStrategyType = (KVS_RETRY_STRATEGY_TYPE) env->CallIntMethod(kvsRetryStrategy, methodId);
```
disa6302, line 345: No need to check `CHK_JVM_EXCEPTION(env);` here?

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
disa6302: You can change this to return STATUS and use the usual CHK methods. In the location where it is called, if it is not status success, you return FALSE or whatever the existing implementation looks like.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+    }
+
+
+CleanUp:
+    if (nullKvsRetryStrategy) {
```
disa6302, line 350: If it is a failure, we do not want to set the values (unless there isnt a cleaner way to do so). The goal is to ensure we do not mask the errors. We need to fail if mandatory conditions are not satisified. In the context of this PR, that would be, having the method IDs set and appropriate members passed wherever relevant.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+        DLOGI("Setting kvsRetryStrategy members to NULL/default values.");
+        pKvsRetryStrategy->pRetryStrategy = NULL;
+        pKvsRetryStrategy->pRetryStrategyConfig = NULL;
+        pKvsRetryStrategy->retryStrategyType = KVS_RETRY_STRATEGY_EXPONENTIAL_BACKOFF_WAIT;
+    }
+}
+
+VOID setKvsRetryStrategyCallbacks(JNIEnv *env, jobject kvsRetryStrategyCallbacks, PKvsRetryStrategyCallbacks pKvsRetryStrategyCallbacks)
```
disa6302, line 358: You can set return type to STATUS and check for status not success in the location it is called in.

src/JNI/com/amazonaws/kinesis/video/producer/jni/Parameters.cpp
```diff
@@ -249,10 +251,194 @@ BOOL setClientInfo(JNIEnv *env, jobject clientInfo, PClientInfo pClientInfo) {
...
+{
+    jmethodID methodId = NULL;
+    jclass cls = NULL;
+    BOOL nullSetCallbacks = FALSE;
```
disa6302, line 362: Can we set this to TRUE instead and set to false wherever relevant? That will simplify this code block. Probably in the previous function too, but I will let you evaluate if that makes sense there.

disa6302 (changes requested):

stefankiesz:
Update from offline discussion: 
- The un-initialized `pClientInfo` members are now null-initialized using a `memset(0)` upon instantiation of the struct.
- The `KvsRetryStrategyCallbacks` are not touched yet in the C-JNI-end. These will need to be setup in the `KinesisVideoClientWrapper` when we add this functionality to the Java SDK.
- The passing of `RetryStrategyType` from Java to here in the JNI was successfully tested using logs.
- In the future, we should consider fatally erroring out when `setClientInfo()` fails (we currently just log a warning).

disa6302 (approved):

hassanctech (approved):
lgtm

PR #1173, opened 2024-05-06
Rework the ReadMe to be more quick-start friendly

stefankiesz:
*What was changed?*
The ReadMe was re-worked to be more quick-start friendly. Missing information was added and made into tables. Typos and grammar issues were addressed.
(More details in the _Overview_ section below.)

*Why was it changed?*
To improve the ReadMe and correct outstanding issues.

*How was it changed?*
By modifying the ReadMe document to address the above issues.

The document was pasted into a spelling and grammar checker.

<br>

### Overview
- Add a collapsible vertical table of contents
- Add a _Quick Start_ section
- Extend samples instructions
- Add missing Windows instructions
- Fix incorrect dependency installation commands
- Add more links to resources and references
  - New links include: AWS KVS getting started page, AWS KVS FAQs page, GStreamer element docs, Java Producer SDK
  -  Also added a link to Producer C in the Key Features section incase user does not need JNI nor GStreamer features
- Make the CMake Arguments section into a table and fix typos in the descriptions
- Make the kvssink parameters into a table, fix missing links, add in aws-region parameter
- Mention and show how to stream from file with the kvssink sample
- To the Quick Start section, add instructions for how/where to view the footage from the samples and link troubleshoot guides
- ...more misc. changes

<br>

### Future Improvements

- Consider splitting the QuickStart instructions for building the SDK into build-with-dependencies a build-using-system-dependencies and demonstrate how to install the system dependencies and the correct versions.
- Improve upon and add to the Debugging/Troubleshooting section - likely should add more links to AWS docs. Specify Java version for JNI. Also should probably add a link to Java SDK and instructions if any are missing on that repo's ReadMe.
- Add installation commands to all the prerequisite packages such as git and cmake. This will likely only fit well if the QuickStart section gets restructured into 3 separate sections for each OS.
- Add a "Running Tests" section to show how to build and run the tests. Include details such as AWS authentication and how resources are managed when tests are running and failing. 

<br>
<br>


README.md
```diff
@@ -10,82 +10,180 @@
...
+## Quick Start
+### Required Tools
+The following packages are required to build the SDK libraries. Using a package manager such as _Homebrew_ (Mac), _APT_ (Linux), and _Chocolatey_ (Windows) is the prefered method of installation.
+* C++ Compiler (GNU or Clang recommended)
+* `git`
+* `CMake`
+* `pkg-config` for _Mac_/_Linux_, `pkgconfiglite` for _Windows_
+* `m4`
+* _Windows_ only: `nasm` and `strawberryperl`
+
```
sirknightj, line 52: Suggest adding the installation commands for these.
```
sudo apt-get update
sudo apt-get install git cmake ...
```

README.md
```diff
@@ -10,82 +10,180 @@
...
+### Build
+#### Prepare a build directory in the newly checked out repository:
 ```bash
 mkdir -p amazon-kinesis-video-streams-producer-sdk-cpp/build
 cd amazon-kinesis-video-streams-producer-sdk-cpp/build
 ```
+#### Configure the build:
 
-If you are building on Windows you need to generate `NMake Makefiles`, you should run `cmake .. -G "NMake Makefiles"`
+_Mac and Linux_
```
sirknightj, line 94: Would suggest splitting this into 2 subsections:
1. Build from scratch (build with dependencies=ON)
2. Build using system libraries (build with dependencies=OFF). For this, you would also need to install the necessary dependencies using system package manager

README.md
```diff
@@ -187,22 +187,28 @@ By default, the samples run in near realtime mode. To set offline mode, set stre
...
-* `-DCOMPILER_WARNINGS` -- Enable all compiler warnings
-* `-DADDRESS_SANITIZER` -- Build with AddressSanitizer
-* `-DMEMORY_SANITIZER` --  Build with MemorySanitizer
-* `-DTHREAD_SANITIZER` -- Build with ThreadSanitizer
-* `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
-* `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
-* `-DBUILD_LOG4CPLUS_HOST` Specify host-name for log4cplus for cross-compilation. Default is OFF.
-* `-DCONSTRAINED_DEVICE` Set the thread stack size to 0.5MB, needed for Alpine builds
-
+You can pass the following CMake options:
```
sirknightj, line 208: You should add "additional". We can pass in things supported by CMake itself, like CMAKE_INSTALL_PREFIX or CMAKE_EXE_LINKER_FLAGS

sirknightj: Things specific to our project

stefankiesz: Makes sense, adding

README.md
stefankiesz: Going to add this to the PR description as a future addition to consider, but I'm not set on doing this quite yet since for the smoothest onboarding experience we would recommend to do build dependencies to ensure a supported version of a dependency library is being used.

stefankiesz: To add to that, it's not as simple as just including the install instructions with the correct versions. Users might already have some of the dependency libraries install with unsupported versions and there would be potential for CMake and/or pkgconfig to link the incorrect ones.

README.md
stefankiesz: This seems like it would be unnecessarily verbose since these are very common packages which are likely already present on users' machines, but I'll add it and see how we feel about it.

stefankiesz: On second thought, it definitely would be to verbose to have all these commands for all 3 OSs, but it should fit well if we restructure the QuickStart section to be 3 separate sections for each OS. Will leave as is for now.

README.md
```diff
@@ -10,213 +10,302 @@
...
+gst-launch-1.0 --gst-version
 ```
 
-If you are building on Windows you need to generate `NMake Makefiles`, you should run `cmake .. -G "NMake Makefiles"`
+<br>
 
-GStreamer and JNI is NOT built by default, if you wish to build both you MUST execute `cmake .. -DBUILD_GSTREAMER_PLUGIN=ON -DBUILD_JNI=TRUE`
+### Download
+```bash
+git clone https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp.git
```
sirknightj, line 84: For faster download, let's clone only the master branch.

README.md
```diff
@@ -10,213 +10,302 @@
...
-* `-DTHREAD_SANITIZER` -- Build with ThreadSanitizer
-* `-DUNDEFINED_BEHAVIOR_SANITIZER` Build with UndefinedBehaviorSanitizer
-* `-DALIGNED_MEMORY_MODEL` Build for aligned memory model only devices. Default is OFF.
-* `-DBUILD_LOG4CPLUS_HOST` Specify host-name for log4cplus for cross-compilation. Default is OFF.
-* `-DCONSTRAINED_DEVICE` Set the thread stack size to 0.5MB, needed for Alpine builds
+<br>
 
-#### To Include JNI
+### Run the Samples
+Included are sample applications for streaming and ingesting to a KVS stream. To stream to KVS, the GStreamer samples use an implemented `appsink` sink element, and the kvssink samples use the custom `kvssink` sink element.
```
sirknightj, line 123: Should add info on credential methods

README.md
```diff
@@ -10,213 +10,302 @@
...
+> More on [GStreamer elements](https://gstreamer.freedesktop.org/documentation/application-development/basics/elements.html?gi-language=c).
 
-JNI examples are NOT built by default.  If you wish to build JNI you MUST add `-DBUILD_JNI=TRUE` when running `cmake`:
+<br>
 
+#### GStreamer appsink Samples
+To stream media from the device's camera and audio sources, create a stream in the [AWS KVS console](https://console.aws.amazon.com/kinesisvideo/home) and run the following sample:
 ```bash
-cmake -DBUILD_JNI=TRUE
+./kvs_gstreamer_audio_video_sample <your-stream-name>
```
sirknightj, line 132: What happens if there's no region set here?

README.md
```diff
@@ -10,213 +10,302 @@
...
 
-To load this plugin set the following environment variables. This should be run from the root of the repo, NOT the `build` directory.
+The intermittent kvssink sample will stream video for 20 seconds, then pause for 40 seconds, and repeat until an interrupt signal is received. To manually adjust the streaming and paused intervals, you can change the `KVS_INTERMITTENT_PLAYING_INTERVAL_SECONDS` and `KVS_INTERMITTENT_PAUSED_INTERVAL_SECONDS` values in the *kvssink_intermittent_sample.cpp* file.
 
 ```bash
-export GST_PLUGIN_PATH=`pwd`/build
-export LD_LIBRARY_PATH=`pwd`/open-source/local/lib
+./kvssink_intermittent_sample <stream-name> <testsrc or devicesrc (optional)>
+```
+Setting the source to `testsrc` will use [videotestsrc](https://gstreamer.freedesktop.org/documentation/videotestsrc/?gi-language=c) and to `devicesrc` will use [autovideosrc](https://gstreamer.freedesktop.org/documentation/autodetect/autovideosrc.html?gi-language=c). By default, kvssink uses "DEFAULT_STREAM" as the stream name, and the sample uses videotestsrc as the source. If a KVS stream with the provided or default name does not exist, the stream will automatically be created.
```
sirknightj, line 185: Should add the stream parameters (2 hours retention)

README.md
```diff
@@ -10,213 +10,302 @@
...
+
+<br>
+
+#### Running Samples in offline mode
+By default, the samples run in near-real-time mode. To set offline mode, set streamInfo.streamCaps.streamingType to `STREAMING_TYPE_OFFLINE`, where, `streamInfo` is of type `StreamInfo`, `streamCaps` is of type `StreamCaps` and `streamingType` is of type `STREAMING_TYPE`.
```
sirknightj, line 190: I think this is confusing, you can explain this a bit better

README.md
```diff
@@ -10,213 +10,302 @@
...
+ 
+### CMake Arguments
+You can pass the following additional CMake options:
+
+| Option	                     | Default       | Description	|
+|:-----------------------------|:-------------:|:-------------|
+| BUILD_GSTREAMER_PLUGIN       | OFF           | Build the kvssink GStreamer plugin
+| BUILD_JNI                    | OFF           | Build C++ wrapper for JNI to expose the functionality to Java/Android
+| BUILD_DEPENDENCIES           | ON            | Build depending libraries from source
+| BUILD_TEST                   | OFF           | Build unit/integration tests, may be useful to confirm support for your device, to run tests:       `./tst/producerTest`
```
sirknightj, line 215: Should mention if credentials & region are required to run the tests & whether the resources get cleaned up or not in case of failing tests.

README.md
```diff
@@ -10,213 +10,302 @@
...
+| CODE_COVERAGE                | OFF           | Enable coverage reporting
+| COMPILER_WARNINGS            | OFF           | Enable all compiler warnings
```
sirknightj, line 217: Why is this off by default?

README.md
```diff
@@ -10,213 +10,302 @@
...
+<br>
+
+## Configuring kvssink
+The kvssink GStreamer element includes the following parameters:
 
+| Parameter	             | Default           | Description	|
+|:-----------------------|:-----------------:|:------------|
+| stream&#x2011;name     | 'DEFAULT_STREAM'  | The name of the destination Kinesis video stream. If a KVS stream with the provided or default name does not exist, the stream will automatically be created.
+| aws&#x2011;region      | 'us-west-2'       | The region specified by this parameter will be used if the AWS_DEFAULT_REGION environment variable is not set.
+| storage&#x2011;size    | 128               | The storage size of the device in megabytes. For information about configuring device storage, see [StorageInfo](https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/producer-reference-structures-producer.html#producer-reference-structures-producer-storageinfo).
```
sirknightj, line 268: Mebibytes/MiB (multiple of 1024)

README.md
```diff
@@ -10,213 +10,302 @@
...
-## Dockerscripts
-* The sample docker scripts for RTSP plugin, raspberry pi and linux can be found in the [Kinesis demos repository](https://github.com/aws-samples/amazon-kinesis-video-streams-demos/tree/master/producer-cpp).
+<br>
 
-## DEBUG
+### Debugging
+
+#### Build Issues
+* If you are running into issues building libcurl on M1 Mac, you can try `brew unlink openssl`.
 * When building the JNI, if you run into a cmake error `Could NOT find JNI (missing: JAVA_INCLUDE_PATH JAVA_INCLUDE_PATH2 JAVA_AWT_INCLUDE_PATH)`, make sure Java is installed and your environment variables are set correctly:
```
sirknightj, line 290: I believe also need java 8 or 11..

README.md
stefankiesz: I tested the difference cloning this on my home network. Cloning the full repo, it takes between 1 and 2 seconds to clone and takes up 14 Mb. With `--single-branch` in the command, it still takes between 1 and 2 seconds to clone and takes up 13 Mb.

I don't think these gains are worth the loss of readily being able to checkout remote branches.

README.md
stefankiesz: DEFAULT_AWS_REGION is used. I added a note of this to the new AWS Authentication section.

README.md
stefankiesz: What's the relevance of that information here? I don't want to crowd all the details here as it is a "quick start" guide and it would be prone to being out of date if we don't update it here if the default parameters ever change. I'd prefer these details live in the AWS docs and not here in the GitHub ReadMe.

README.md
stefankiesz: This I just copied over from the existing ReadMe, but looking at it now, we don't expose a StreamInfo object in the samples. Instead, the `streaming_type` variable should be modified to change the streaming type. However, I don't see why this would ever need to be done if the sample already changes it to offline mode when streaming from file. I'll remove this section entirely if you agree.

README.md
stefankiesz: I feel this would be too verbose for this table. I will add including a "Running Tests" section to the PR description as a potential future addition where these details would fit in well.

README.md
stefankiesz: I wouldn't know as this was likely a decision made long ago, but I'd imaging it was to reduce build log verbosity.

Also, the COMPILER_WARNINGS flag does nothing in PIC, Producer C, and Producer CPP, yet we have it as an option in all 3 SDKs. It's not a built-in option as far as what I can see from a quick search.

Let's just remove this from the table? ...and from the CMakeLists in a separate PR.

README.md
stefankiesz: Nice catch

README.md
stefankiesz: Improving upon the Debugging section is already in the PR description as an action item, this will fall in that scope as it will need testing and verification.

README.md
sirknightj: I think that's a good idea. Having a section dedicated to how to run the tests and the sanitizers and view the results.

README.md
sirknightj: > This seems like it would be unnecessarily verbose since these are very common packages which are likely already present on users' machines, but I'll add it and see how we feel about it.

In the case that it's not or if I want to verify for myself that it is installed? This information seems quite important to have.

README.md
```diff
@@ -10,213 +10,320 @@
...
+brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly gst-libav
+```
+_Linux_
+```bash
+sudo apt install libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev gstreamer1.0-plugins-base-apps gstreamer1.0-plugins-bad gstreamer1.0-plugins-good gstreamer1.0-plugins-ugly gstreamer1.0-tools
+```
+_Windows_
+```bat
+choco install gstreamer --version=1.22.8
+choco install gstreamer-devel --version=1.22.8
```
sirknightj, line 72: How come the gstreamer version is only specified in the windows? Does mac and linux requires certain gstreamer versions? Is this the only supported gstreamer version for windows?

README.md
```diff
@@ -10,213 +10,320 @@
...
+```
 
-Prepare a build directory in the newly checked out repository:
+<br>
 
+### Build
+#### Prepare a build directory in the newly checked out repository:
 ```bash
 mkdir -p amazon-kinesis-video-streams-producer-sdk-cpp/build
 cd amazon-kinesis-video-streams-producer-sdk-cpp/build
```
sirknightj, line 93: Don't we need to clone it in the `C:/` for windows due to file path length restrictions?

Also can we adjust this to clone only the master branch?

sirknightj: If you could add a note for that one, that would be great.

README.md
sirknightj: Sure thing, we can re-evaluate it at a later date.

README.md
sirknightj: Yeah just remove it. It's already explained in docs/structures.md

README.md
stefankiesz: Not needed when building Windows locally, we move it in github actions because it has it nested in `amazon-kinesis-video-streams-producer-sdk-cpp/amazon-kinesis-video-streams-producer-sdk-cpp` which is not going to be the case when cloning locally.

README.md
stefankiesz: Likely not the only supported version, but for the windows build, we have relied on using the pkg-config library bundled with the gstreamer installation. We've found that some versions of gstreamer do not have the pkg-config library. I believe we may be able to switch to just using pkg-config-lite, installed via choco, then we can remove this version specifier.

README.md
stefankiesz: Added the installation steps 👍

PR #1174, opened 2024-05-07
Add a parameter to kvssink to enable tags needed for S3 image delivery

akadesilva:
*Issue #, if available:*

*Description of changes:*

A new parameter 'img_gen' to enable or disable MKV tags needed to activate S3 image delivery
img_gen=1 sets the tags for every keyframe
img_gen=0 does not add the tags


stefankiesz:
@akadesilva thanks a lot for the contribution! Image generation will be great a great feature for kvssink to support.

I'll take a closer look at the use cases for image generation, but my immediate thoughts are that we might not want to generate an image for every keyframe, but instead have a custom event that can be emitted to kvssink which will react by marking the next received keyframe for image generation. This would be implemented similarly to the [fragment metadata feature](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1122/files).

My thinking is with this, we can have, say, a human detector element send a message to the application whenever it detects humans in frames on the way down the pipeline to kvssink, and then the application can react to this by sending a message to kvssink to save the next received keyframe as an image - or however many keyframes the application would like to keep generating images for after such an event occurs.

sirknightj:
Suggestion to make it more versatile in addition to the event listener is to adjust it to become the gap between keyframes.

Meaning:

0 = OFF (Default)
1 = Every keyframe
2 = Every other keyframe
3 = Every 3rd keyframe
... and so on

stefankiesz:
This sounds great to me, @sirknightj. So let's have this parameter be an int specifying keyframes apart and add the functionality in kvssink to track which keyframe we are on to determine whether to generate an image. And we can eventually also do the message-based approach I suggested as a separate feature.

akadesilva:
Thanks for the comments! I am using this since I have an application where I need to get images using the S3 image delivery feature in KVS. I have a web ui which is setting the sampling interval (typically 1-5s) via the update image generation api call to KVS. The generated images are used in a ML application. 
I am happy to add the add the feature suggested by @sirknightj to make it more flexible. Using the event message looks tricky since I'm not sure how to send a message to kvssink when using gst-launch in the command line. Happy to look into it further if you can give me few pointers

sirknightj:
Let's keep this PR just for adding this s3 tag interval and we can do the message processing in subsequent PR.

The outline for this message processing feature would go like this:
1. In a new sample, create a [new thread](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/93f35cf9c5f369ab1eb778996bf35ca5e566ebaa/samples/kvssink_gstreamer_sample.cpp#L783) running a loop that [emits a new s3 image delivery tag event](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/93f35cf9c5f369ab1eb778996bf35ca5e566ebaa/src/gstreamer/gstkvssink.h#L215-L222).
2. Add a new function to the [event handler](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/93f35cf9c5f369ab1eb778996bf35ca5e566ebaa/src/gstreamer/gstkvssink.cpp#L1154) that processes this new s3 image delivery tag event.
3. Stop + cleanup the thread when `kvssink` is torn down.

For this PR:
1. Please tweak the parameter to an `int` as discussed above
2. Change the target branch to the `develop` branch

After that, we can get this merged in. Thanks for your contribution!

akadesilva:
I have updated the feature for kvs sink to generate every N number of keyframes when img_gen=N. However I noticed that there was a new PR https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1176 which adds this feature across the board. So I don't think this PR is needed any more

niyatim23:
> I have updated the feature for kvs sink to generate every N number of keyframes when img_gen=N. However I noticed that there was a new PR #1176 which adds this feature across the board. So I don't think this PR is needed any more

Hi @akadesilva, apologies for the same. This feature was sitting in my branch since January this year. I was able to clean it up and get it out now. We really appreciate you taking the time to contribute

akadesilva:
No worries. Thanks for adding it across the board. Makes things so much more stream lined

PR #1176, opened 2024-05-12
[FEATURE] Kvssink image-generation

niyatim23:
*Issue #, if available:*

*Description of changes:*


sirknightj:
As discussed in the other PR, please adjust to every N keyframes to make this feature more flexible.

Also, please edit PR title, description, GStreamer parameter description, etc to clearly mark that this feature is explicitly for __s3 image delivery__. The s3 delivery mkv simple tag is not needed for image generation through GetImages, and thus, "image generation" is already supported in its current state.

For completeness, can you also share an outline of the procedure you used to verify the tag was added and a screenshot of some media with the tag.

niyatim23:
> As discussed in the other PR, please adjust to every N keyframes to make this feature more flexible.
> 
> Also, please edit PR title, description, GStreamer parameter description, etc to clearly mark that this feature is explicitly for **s3 image delivery**. The s3 delivery mkv simple tag is not needed for image generation through GetImages, and thus, "image generation" is already supported in its current state.
> 
> For completeness, can you also share an outline of the procedure you used to verify the tag was added and a screenshot of some media with the tag.

Is there a customer who requested the N key frames feature? Where did this request originate? I don't see that in producer c. This feature has been called "image-generation" in the past so I would like to stick to the name to keep it uniform. I verified that the tag is added with mkvinfo:
```
|+ Tags
| + Tag
|  + Simple
|   + Name: AWS_KINESISVIDEO_IMAGE_GENERATION
|   + String
|+ Cluster
| + Cluster timestamp: 476564:53:17.855000000
| + Cluster position: 0
| + Simple block: key, track number 1, 1 frame(s), timestamp 476564:53:17.855000000
|  + Frame with size 5666
|+ Tags
| + Tag
|  + Simple
|   + Name: AWS_KINESISVIDEO_IMAGE_GENERATION
|   + String
|+ Tags
| + Tag
|  + Simple
|   + Name: AWS_KINESISVIDEO_END_OF_FRAGMENT
|   + String
```

hassanctech:
Discussed offline waiting for a change here to trigger a call to the put event metadata API based on receiving GST_EVENT_CUSTOM_DOWNSTREAM similar to the put fragment metadata API.

sirknightj:
I will push for keeping consistent with the docs and calling this [**image delivery**](https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/gs-s3Delivery.html) to avoid confusion with the other feature getimages https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/gs-getImages.html

Also, please update the readme as there is no information about how to use it.

src/KinesisVideoStream.cpp
```diff
@@ -185,6 +185,16 @@ bool KinesisVideoStream::putFragmentMetadata(const std::string &name, const std:
     return true;
 }
 
+bool KinesisVideoStream::putEventMetadata(uint32_t event, PStreamEventMetadata pStreamEventMetadata){
+    STATUS status = ::putKinesisVideoEventMetadata(stream_handle_, event, pStreamEventMetadata);
+    if (STATUS_FAILED(status)) {
```
unicornss, line 190: any example scenario(s) when this could happen?

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -333,6 +349,14 @@ static void put_metadata(GstElement* element) {
     }
 }
 
+static void deliver_images(GstElement* element) {
+    if (!put_event_metadata(element, STREAM_EVENT_TYPE_NOTIFICATION | STREAM_EVENT_TYPE_IMAGE_GENERATION, NULL)) {
```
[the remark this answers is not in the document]
niyatim23: I'm following what we have in other samples: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/blob/master/samples/KvsAudioVideoStreamingSample.c#L78

src/KinesisVideoStream.cpp
niyatim23: That's the status code returned by the C function, so it will be all possible status codes that can be returned from [here](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/65e38dac9b30523d43a57bc009d679e627b58d9a/src/client/src/Client.c#L972) that indicate failure

src/gstreamer/gstkvssink.cpp
```diff
@@ -122,6 +122,10 @@ GST_DEBUG_CATEGORY_STATIC (gst_kvs_sink_debug);
 #define KVS_ADD_METADATA_VALUE "value"
 #define KVS_ADD_METADATA_PERSISTENT "persist"
 
+#define KVS_ADD_EVENT_METADATA_G_STRUCT_NAME "kvs-add-event-metadata"
+#define KVS_ADD_EVENT_METADATA_EVENT "event"
```
unicornss, line 126: Does gst-inspect-1.0 kvssink print info / actions about  new capability?

samples/kvssink_gstreamer_sample.cpp
```diff
@@ -795,7 +819,8 @@ int gstreamer_init(int argc, char *argv[], CustomData *data) {
 
     // Create a GStreamer timer to generate and put fragment metadata tags every 2 seconds
```
unicornss, line 820: Is the purpose of this sample to show sending an image event every two seconds? Is the number of images specified at the level of the stream?

samples/kvssink_gstreamer_sample.cpp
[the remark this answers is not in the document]
unicornss: We could remove the notification here as the comment there refers to both notifications and images, but this is a _deliver_images_ function.

src/KinesisVideoStream.h
```diff
@@ -83,6 +83,18 @@ class KinesisVideoStream {
      */
     bool putFragmentMetadata(const std::string& name, const std::string& value, bool persistent = true);
 
+    /*
+     * Inserts a KVS event(s) accompanied by optional metadata (key/value string pairs) into the stream.
+     * Multiple events can be submitted at once by using bitwise OR of event types, or multiple calls of this
+     * function with different unique events.
+     * @param 1 uint32_t - the type of event(s), a value from STREAM_EVENT_TYPE enum. If
```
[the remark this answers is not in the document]
niyatim23: [Here](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/65e38dac9b30523d43a57bc009d679e627b58d9a/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L973-L985). There can only be 10 metadata per fragment

samples/kvssink_gstreamer_sample.cpp
niyatim23: The purpose is to show that this can be done repeatedly per fragment and not just once. Since this is just an example, we put the event until we reach the limit

src/gstreamer/gstkvssink.cpp
niyatim23: It doesn't because this feature cannot be used from the command line because the function needs to be configured

jameselsey:
@niyatim23 Do you have any updates for this PR? This feature would be really useful! 👍

PR #1193, opened 2024-06-20
Allow using prebuilt libkvscproducer through pkg-config

hacker1024:
*Description of changes:*

At the moment, the use of `libkvscproducer` does not respect `-DBUILD_DEPENDENCIES`. This PR allows it to be found via `pkg-config` instead of built from source.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,17 +94,24 @@ set(BUILD_COMMON_LWS
 set(BUILD_COMMON_CURL
     TRUE
     CACHE BOOL "Build ProducerC with CURL Support" FORCE)
-set(DEPENDENCY_DOWNLOAD_PATH ${CMAKE_CURRENT_SOURCE_DIR}/dependency)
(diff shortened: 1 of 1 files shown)


sirknightj:
Hi @hacker1024, thanks for your contribution.

We want to avoid breaking backwards compatibility. The behavior we want to make sure doesn't get broken, is that when running `cmake .. -DBUILD_DEPENDENCIES=OFF` after a fresh git clone, it still builds C producer (for a new user who doesn't have C producer cloned or built). I would suggest adding a new flag which build producer-c defaulting to ON to maintain this behavior.

In your PR description, could you also add how you tested it (which scenarios/commands you ran)?

Finally, could you change the target branch to `develop`? We'll merge develop into master branch during the next release cycle.

Looking forward to these enhancements!

sirknightj:
Might be related: https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1115

Issue #1225, opened 2025-01-22
[BUG]GStreamer unit tests with Segmentation Fault


shotamishima:
**Logging**
```
Unexpected critical/warning: gst_pad_get_peer: assertion 'GST_IS_PAD (pad)' failed

Stack trace:
0   libgstreamer-1.0.0.dylib            0x0000000102f3db8c gst_debug_get_stack_trace + 52
1   libgstcheck-1.0.0.dylib             0x0000000102c26d3c gst_check_log_critical_func + 144
2   libglib-2.0.0.dylib                 0x0000000103063dd0 g_logv + 828
3   libglib-2.0.0.dylib                 0x0000000103063a24 g_log + 28
4   libglib-2.0.0.dylib                 0x0000000103063814 g_return_if_fail_warning + 32
5   libgstreamer-1.0.0.dylib            0x0000000102f4cc8c gst_pad_get_peer + 128
6   libgstcheck-1.0.0.dylib             0x0000000102c27448 gst_check_teardown_pad_by_name + 24
7   gstkvsplugintest                    0x0000000102b71a2c _ZL23kvsproducersinkteststopi + 172
8   libgstcheck-1.0.0.dylib             0x0000000102c32eb0 srunner_run_tagged + 1780
9   libgstcheck-1.0.0.dylib             0x0000000102c287e0 gst_check_run_suite + 116
10  gstkvsplugintest                    0x0000000102b72e8c main + 56
11  dyld                                0x00000001950d0274 start + 2840



Unexpected critical/warning: gst_pad_get_peer: assertion 'GST_IS_PAD (pad)' failed

Stack trace:
0   libgstreamer-1.0.0.dylib            0x0000000102f3db8c gst_debug_get_stack_trace + 52
1   libgstcheck-1.0.0.dylib             0x0000000102c26d3c gst_check_log_critical_func + 144
2   libglib-2.0.0.dylib                 0x0000000103063dd0 g_logv + 828
3   libglib-2.0.0.dylib                 0x0000000103063a24 g_log + 28
4   libglib-2.0.0.dylib                 0x0000000103063814 g_return_if_fail_warning + 32
5   libgstreamer-1.0.0.dylib            0x0000000102f4cc8c gst_pad_get_peer + 128
6   libgstcheck-1.0.0.dylib             0x0000000102c27448 gst_check_teardown_pad_by_name + 24
[...]

PR #1226, opened 2025-01-24
kvssink CI enhancements!

sirknightj:
*Issue #, if available:*
#1225

*Description of changes:*

## 1. Fixing the unit tests with kvssink.

Unit tests were reported broken in the issue above. Due to a combination of two combined issues:
1. Implicit bad gchar* (aka char*) -> std::string conversions. If a nullptr is passed into the std::string constructor, a segfault will occur.
2. Missing pads in the unit tests

In KVS client & stream initialization, the `kvssink->content_type` (gchar*) field is obtained from the element directly to the left (pad). However, in the unit tests, the pad was not constructed or attached. As such, that `kvssink->content_type` is a nullptr. When it gets passed into the StreamDefinition constructor, it gets implicitly cast into the `std::string`.

Note that the null inputs case is _only_ possible through programatic use of GStreamer and kvssink. When using `gst-launch-1.0` or `gst_parse_launch()`, properties can't be set to null, since the kvssink properties are configured with default values. Hence, this effort will assist developers using `kvssink` programatically.

To fix this:
1. The pads were constructed and linked in the unit tests.
2. Added null checks to the C- char* strings before converting them to C++ strings.
3. Added new unit tests to check that kvssink exits gracefully if the gchar* fields are nullptrs.

### 2.1. Adding a session token parameter to the unit tests
* Since we need to use temporary credentials, the gstreamer unit tests now have the session token logic added.

### 2.2. Running the repaired gstreamer tests automatically for every pull request
* Using the docker images from the AWS ECR gallery to avoid random github actions runner updates from breaking the builds.

## Testing:
* Ran it on my local setup to check that the tests pass. Tried both with session token and without.
* The new error messages show up as part of the GStreamer logs. Run `GST_DEBUG=4 ./tst/gstkvsplugintest`.
```
0:00:00.035717708 33125 0x600000e8c5d0 WARN                 kvssink gstkvssink.cpp:1672:gst_kvs_sink_change_state:<kvssink> error: Failed to init kvs producer. Error: Assertion failed: kvssink->stream_name is unexpectedly NULL!

0:00:00.031656167 25576 0x600000f5c960 WARN                 kvssink gstkvssink.cpp:1672:gst_kvs_sink_change_state:<kvssink> error: Failed to init kvs producer. Error: Assertion failed: kvssink->content_type is unexpectedly NULL!
```

## 3. Verifying the debug dump dir and kvssink on different flavors (Windows, mac, linux).
- Run kvssink with videotestsrc for 15 seconds with the KVS_DEBUG_DUMP_DATA_FILE_DIR set. Then, run `mkvinfo` on the output MKV file to verify that it's generated.

> [!NOTE]
> The valgrind checks currently work but have been commented out as they are failing. Fixing the memory leak in the kvssink _unit tests_ can come as part of a future development.


.github/workflows/ci.yml
```diff
@@ -115,131 +115,7 @@ jobs:
...
-          role-session-name: ${{ secrets.AWS_ROLE_SESSION_NAME }}
-          aws-region: ${{ secrets.AWS_REGION }}
-          role-duration-seconds: 10800
-      - name: Run tests
-        run: |
-          cd build
-          ulimit -c unlimited -S
-          timeout --signal=SIGABRT 60m ./tst/producerTest
-
-  undefined-behavior-sanitizer:
```
stefankiesz, line 155: What was the reason for removing these 2 tests from here?

src/gstreamer/gstkvssink.cpp
```diff
@@ -423,6 +435,13 @@ void create_kinesis_video_stream(GstKvsSink *kvssink) {
             break;
     }
 
+    // StreamDefinition takes in C++ strings, check the gchar* for nullptr before constructing
+    if (kvssink->stream_name == NULL || kvssink->content_type == NULL ||
+        kvssink->user_agent == NULL || kvssink->kms_key_id == NULL ||
+        kvssink->codec_id == NULL || kvssink->track_name == NULL) {
+        throw runtime_error(string(INTERNAL_CHECK_PREFIX) + " A string was unexpectedly NULL!");
```
stefankiesz, line 442: Please add more context to this log such as "A StreamDefinition string member was..."

.github/workflows/ci.yml
sirknightj: I think I was in the middle of migrating the sanitizer tasks to their own file, will revert.

src/gstreamer/gstkvssink.cpp
sirknightj: I extracted the logic to a macro to print out the exact member that was null.

sirknightj: Checked locally:
```
0:00:00.039928791  3604 0x6000011501e0 WARN                 kvssink gstkvssink.cpp:1667:gst_kvs_sink_change_state:<kvssink> error: Failed to init kvs producer. Error: Assertion failed: kvssink->content_type is unexpectedly NULL!
```

sirknightj:
Rpi checks failing (expected) and everything else passing

src/gstreamer/gstkvssink.cpp
```diff
@@ -476,6 +492,15 @@ bool kinesis_video_stream_init(GstKvsSink *kvssink, string &err_msg) {
             create_kinesis_video_stream(kvssink);
             break;
         } catch (runtime_error &err) {
+
+            // Don't retry if the error is an internal error
+            if (STRNCMP(INTERNAL_CHECK_PREFIX, err.what(), STRLEN(INTERNAL_CHECK_PREFIX)) == 0) {
```
unicornss, line 497: can we move this up?

   ostringstream oss;
    oss << "Failed to create stream. Error: " << err.what();
    err_msg = oss.str();

tst/gstreamer/gstkvstest.cpp
```diff
@@ -86,6 +111,49 @@ GST_START_TEST(kvsproducersinkteststop)
     }
 GST_END_TEST;
 
+GST_START_TEST(check_stream_name_null_handled)
```
unicornss, line 114: check_kvssink_null_stream_name_fails init play

tst/gstreamer/gstkvstest.cpp
```diff
@@ -86,6 +111,49 @@ GST_START_TEST(kvsproducersinkteststop)
...
+        // Test - Initialization of the client & stream occurs here
+        fail_unless_equals_int(GST_STATE_CHANGE_FAILURE, gst_element_set_state(kinesisvideoproducersink, GST_STATE_PLAYING));
+
+        // Teardown
+        fail_unless_equals_int(GST_STATE_CHANGE_SUCCESS, gst_element_set_state(kinesisvideoproducersink, GST_STATE_NULL));
+        gst_check_teardown_element(kinesisvideoproducersink);
+    }
+GST_END_TEST;
+
+GST_START_TEST(check_no_pads_content_type_set_correctly)
```
unicornss, line 135: check_kvssink_no_pads_fails_to_play ?

tst/gstreamer/gstkvstest.cpp
```diff
@@ -217,20 +292,47 @@ GST_START_TEST(check_playing_to_paused_and_back_to_playing)
...
+    accessKey = GETENV(ACCESS_KEY_ENV_VAR);
+    secretKey = GETENV(SECRET_KEY_ENV_VAR);
+    sessionToken = GETENV(SESSION_TOKEN_ENV_VAR);
+
+    accessKey = accessKey ? accessKey : "";
+    secretKey = secretKey ? secretKey : "";
+    sessionToken = sessionToken ? sessionToken : "";
+
+    // Check if required environment variables are set
+    if (accessKey[0] == '\0' || secretKey[0] == '\0') {
```
unicornss, line 324: ||sessionToken[0] == '\0'

tst/gstreamer/gstkvstest.cpp
```diff
@@ -217,20 +292,47 @@ GST_START_TEST(check_playing_to_paused_and_back_to_playing)
...
+
+        gst_pad_set_active (srcpad, FALSE);
+
         cleanup_kinesisvideoproducersink(pElement);
     }
 GST_END_TEST;
 
+GST_START_TEST(test_check_credentials)
+    {
+        ck_abort_msg("Required environment variables (ACCESS_KEY and/or SECRET_KEY) are not set");
```
unicornss, line 304: can it be extended to catch specifics instead of AK OR SK?

unicornss (approved):

tst/gstreamer/gstkvstest.cpp
sirknightj: Yes

tst/gstreamer/gstkvstest.cpp
sirknightj: Not checking session token here since it can be empty string if permanent credentials are used. Added a comment.

We currently don't do further validation (e.g. regex validation) on the access key, secret key, session token variables in the rest of the SDK. We can look into adding that in the future.

stefankiesz (approved):

PR #1243, opened 2025-03-25
Build optimizations autoconf automake

sirknightj:
*Issue #, if available:*
- N/A

*Description of changes:*
- See #1151 for the original details
- Use Cmake directly to build log4cplus instead of building and using automake and autoconf
- Remove unnecessary checks for openssl 
- Remove dependencies on C producer in the sample (with the systemCurrentTime() function) to use Chrono instead

Extra:
- Shorten the time that the role is assumed for (CI doesn't need that long of a credentials)

*Testing*
- Checked locally on my mac and also on my fork.


sirknightj:
The host platform changes aren't backwards compatible for cross-compiling cases. We can revisit this in the future.

PR #1292, opened 2026-04-21
Close stale issues workflow

vabhasin:
*Issue #, if available:*
N/A

*What was changed?*
Add a github workflow using `aws-actions/stale-issue-cleanup@v7` for this workflow.
The workflow runs on all issues that are tagged with label `pending-response` (waiting on author's response).
- If the issue has seen no activity for 7 days, it will be tagged with label `closing-soon`
- Following that, if the issue still doesn't see any activity for 3 more days it will be closed (reason: staleness)

*Why was it changed?*
- Automate closing of stale issues

*How was it changed?*
- Github actions

- Verification of the workflow can only be done when merging in the default(master) branch. 
- Can be improvised later based on how it runs

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Close stale issues
     steps:
-      - uses: aws-actions/stale-issue-cleanup@v3
+      - uses: aws-actions/stale-issue-cleanup@v7
         with:
           ancient-issue-message: This is a very old issue. We encourage you to check if this is still an issue in the latest release and if you find that this is still a problem, please feel free to open a new one.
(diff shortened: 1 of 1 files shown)


.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,38 @@
+name: Close Stale Issues
+
+# Controls when the action will be run
+# Run every day at 00:00
+on:
```
sirknightj, line 5: Recommend to add a condition to skip in non-awslab repos (don't run on forks)

vabhasin: Looks like workflows are disabled by default on forked repos. (https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-workflows)
> When a public repository is forked, scheduled workflows are disabled by default

sirknightj (approved):

Ahhorse1 (approved):

PR #1298, opened 2026-07-15
make gst pipeline resilient to transient recoverable errors

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- In `gst_kvs_sink_handle_buffer`, stream error handling no longer unconditionally terminates the GStreamer pipeline. A stream error is now fatal only when `restart-on-error` is disabled OR when running in offline mode. In realtime mode with `restart-on-error` enabled (the default), the stream is reset and the pipeline continues running.
- In `KvsSinkStreamCallbackProvider::streamErrorReportHandler`, added `!IS_RETRIABLE_ERROR(status_code)` to the filter that latches `stream_status` and emits the `stream-error` signal, so retriable control-plane errors are no longer treated as fatal.

*Why was it changed?*
- Previously, the error check `IS_OFFLINE_STREAMING_MODE(...) || !IS_RETRIABLE_ERROR(stream_status)` would kill the entire pipeline for any error outside the narrow `IS_RETRIABLE_ERROR` set. Data-plane errors received via ACKs during streaming would terminate a long-running realtime pipeline unnecessarily. For a 24/7 camera, a single transient backend error should not tear down the whole GStreamer pipeline.
- Additionally, the recovery path was calling `GST_ELEMENT_ERROR`, which posts a fatal `GST_MESSAGE_ERROR` to the pipeline bus. Even though the C code continued (reset stream, reset status), the bus message caused the application/`gst-launch` to shut the pipeline down, which defeats the entire purpose.
- The error-report handler filtered `!IS_RECOVERABLE_ERROR && !IS_RETRIABLE_COMMON_LIB_ERROR` but was missing `!IS_RETRIABLE_ERROR`, making it inconsistent with the KVS GStreamer samples (`kvs_gstreamer_sample.cpp`, `kvs_gstreamer_audio_video_sample.cpp`) and the continuous-retry callback, both of which treat `IS_RETRIABLE_ERROR` as non-fatal. Retriable control-plane errors (describe/create/putStream/getToken/getEndpoint) are already retried by PIC's state machine and the continuous-retry `resetStream`, so latching them as fatal `stream_status` caused multiple redundant resets or unwanted pipeline termination.

*How was it changed?*
- Replaced the fatal-condition check with `!kvssink->restart_on_error || IS_OFFLINE_STREAMING_MODE(kvssink->streaming_type)`. This reuses the existing `restart-on-error` property (default TRUE) so users have explicit control over recovery behavior.
- Fatal path (offline, or `restart-on-error=false`): posts `GST_ELEMENT_ERROR` and returns `GST_FLOW_ERROR` to terminate the pipeline.
- Recovery path (realtime + `restart-on-error=true`): posts `GST_ELEMENT_WARNING` (non-fatal, logged and surfaced to the app but does not stop the pipeline), calls `resetStream()` to flush the producer buffer and re-establish the connection, then resets `stream_status` and continues.
- Added `!IS_RETRIABLE_ERROR(status_code)` to `streamErrorReportHandler` so only genuinely fatal errors (neither retriable, recoverable, nor common-lib-retriable) latch `stream_status` and raise the `stream-error` signal, aligning kvssink with the sample handlers and preventing double-handling of errors the lower layers already recover from.

- Built successfully on macOS (clang) and Ubuntu 22.04 (gcc) with `-DBUILD_GSTREAMER_PLUGIN=ON -DBUILD_TEST=ON`.
- Existing tests pass
- Simulated 403 and saw that PIC/Producer layer retries, kvssink logs error but doesn't terminate the pipeline. With `restart-on-error` enabled, kvssink terminates the pipeline cleanly.

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1305,15 +1305,17 @@ gst_kvs_sink_handle_buffer (GstCollectPads * pads,
     info.data = NULL;
 
     if (STATUS_FAILED(stream_status)) {
+        GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL),
+            ("Stream error occurred. Status: 0x%08x", stream_status));
         // in offline case, we cant tell the pipeline to restream the file again in case of network outage.
         // therefore error out and let higher level application do the retry.
-        if (IS_OFFLINE_STREAMING_MODE(kvssink->streaming_type) || !IS_RETRIABLE_ERROR(stream_status)) {
+        if (IS_OFFLINE_STREAMING_MODE(kvssink->streaming_type)) {
             // fatal cases
-            GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL),
-                           ("[%s] Stream error occurred. Status: 0x%08x", kvssink->stream_name, stream_status));
+            GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL), ("Stop stream in offline mode"));
             ret = GST_FLOW_ERROR;
             goto CleanUp;
         } else {
+            GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL), ("Reset stream"));
             // resetStream, note that this will flush out producer buffer
             data->kinesis_video_stream->resetStream();
             // reset state
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1312,14 +1312,14 @@ gst_kvs_sink_handle_buffer (GstCollectPads * pads,
         if (!kvssink->restart_on_error || IS_OFFLINE_STREAMING_MODE(kvssink->streaming_type)) {
             // fatal case: post a fatal ERROR to the bus so the pipeline terminates
             GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL),
-                ("Stream error occurred. Status: 0x%08x", stream_status));
+                ("[%s] Stream error occurred. Status: 0x%08x", kvssink->stream_name, stream_status));
             ret = GST_FLOW_ERROR;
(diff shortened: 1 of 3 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -1305,15 +1305,17 @@ gst_kvs_sink_handle_buffer (GstCollectPads * pads,
     info.data = NULL;
 
     if (STATUS_FAILED(stream_status)) {
+        GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL),
```
sirknightj: Perhaps we can make use of this parameter to make it configurable:


```
            kvssink->restart_on_error,// SDK will restart when error happens
```

sirknightj (approved):

src/gstreamer/gstkvssink.cpp
```diff
@@ -1305,15 +1305,21 @@ gst_kvs_sink_handle_buffer (GstCollectPads * pads,
...
-            // fatal cases
+        // A stream error is fatal to the pipeline when:
+        //   - restart-on-error is disabled (user opted out of recovery), OR
+        //   - we are in offline mode: we can't re-stream a file after a network
+        //     outage, so let the higher level application handle the retry.
+        if (!kvssink->restart_on_error || IS_OFFLINE_STREAMING_MODE(kvssink->streaming_type)) {
+            // fatal case: post a fatal ERROR to the bus so the pipeline terminates
             GST_ELEMENT_ERROR (kvssink, STREAM, FAILED, (NULL),
-                           ("[%s] Stream error occurred. Status: 0x%08x", kvssink->stream_name, stream_status));
+                ("Stream error occurred. Status: 0x%08x", stream_status));
```
Ahhorse1: Why are we removing the stream name from the log?

Ahhorse1 (approved):

PR #1299, opened 2026-07-16
Add fragment boundary presentation timestamps logs

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Added fragment boundary logging to kvssink. On each keyframe, the previous fragment's start/end timestamps and the gap to the next fragment are logged. Also logs the PTS baseline (first_pts, producer_start_time) once at stream start.

*Why was it changed?*
- To add debugging ability  for possible fragment overlap. Currently there's no visibility into fragment-level timestamps within kvssink. These logs show the exact gap (or negative overlap) between consecutive fragments, making root-cause analysis straightforward.

*How was it changed?*
- Added 5 fields to `_KvsSinkCustomData` in `gstkvssink.h`: `fragment_start_pts`, `fragment_max_pts`, `fragment_count`, `fragment_start_pts_ns`, `fragment_max_pts_ns`.
- In `put_frame()`, after `statusPutFrame`, track fragment boundaries on keyframes. Log the previous fragment's epoch-ms timestamps and the gap to the current keyframe. Also log raw GStreamer running-time PTS for correlation with GStreamer debug output.
- In `gst_kvs_sink_handle_buffer`, log the PTS baseline once when `producer_start_time` is first set.
- All logs are at DEBUG level, emitted only on keyframes (~once per 1.5-10s depending on encoder settings). Zero overhead when log level is above DEBUG.

- Ran end-to-end with `kvssink_gstreamer_sample` using videotestsrc at 30fps. Verified log output shows expected ~33ms gap between fragments and correct epoch-ms timestamps matching KVS console playback timeline.
- Example output:

```
[INFO ] [16-07-2026 00:30:19:733.117 UTC]  kvssink PTS baseline: first_pts=0ns producer_start_time=1784160945528000000ns
[DEBUG] [16-07-2026 00:30:30:191.785 UTC] Fragment #7 start=1784161828733 end=1784161830199 next_start=1784161830233 gap=34ms
[DEBUG] [16-07-2026 00:30:30:191.804 UTC] Fragment #7 gst_start=9171822417 gst_end=10638489083 gst_next_start=10671822417ns
```

### DIFF
--- a/src/gstreamer/gstkvssink.cpp
+++ b/src/gstreamer/gstkvssink.cpp
@@ -1270,6 +1270,39 @@ put_frame(std::shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len,
 
     create_kinesis_video_frame(&frame, pts, dts, flags, frame_data, len, track_id, index);
     put_frame_status = data->kinesis_video_stream->statusPutFrame(frame);
+
+    // Log fragment boundaries: keyframe = new fragment start
+    if (CHECK_FRAME_FLAG_KEY_FRAME(flags)) {
+        if (data->fragment_count > 0) {
+            // Epoch ms timestamps (matching KVS backend's view)
+            uint64_t prev_start_ms = data->fragment_start_pts * DEFAULT_TIME_UNIT_IN_NANOS / 1000000;
+            uint64_t prev_end_ms = data->fragment_max_pts * DEFAULT_TIME_UNIT_IN_NANOS / 1000000;
+            uint64_t curr_start_ms = frame.presentationTs * DEFAULT_TIME_UNIT_IN_NANOS / 1000000;
+            LOG_DEBUG("Fragment #" << data->fragment_count
+                     << " start=" << prev_start_ms
+                     << " end=" << prev_end_ms
+                     << " next_start=" << curr_start_ms
+                     << " gap=" << (int64_t)(curr_start_ms - prev_end_ms) << "ms");
+            // Raw GStreamer PTS (running time ns)
+            uint64_t offset_ns = data->producer_start_time - data->first_pts;
+            LOG_DEBUG("Fragment #" << data->fragment_count
+                     << " gst_start=" << (data->fragment_start_pts_ns - offset_ns)
+                     << " gst_end=" << (data->fragment_max_pts_ns - offset_ns)
+                     << " gst_next_start=" << (static_cast<uint64_t>(pts.count()) - offset_ns) << "ns");
+        }
+        data->fragment_start_pts = frame.presentationTs;
+        data->fragment_max_pts = frame.presentationTs;
+        data->fragment_start_pts_ns = static_cast<uint64_t>(pts.count());
+        data->fragment_max_pts_ns = static_cast<uint64_t>(pts.count());
+        data->fragment_count++;
+    }
+
+    // Track max PTS across all frames in the current fragment
+    if (frame.presentationTs > data->fragment_max_pts) {
+        data->fragment_max_pts = frame.presentationTs;
+        data->fragment_max_pts_ns = static_cast<uint64_t>(pts.count());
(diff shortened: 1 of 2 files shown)


src/gstreamer/gstkvssink.cpp
```diff
@@ -1270,6 +1270,39 @@ put_frame(std::shared_ptr<KvsSinkCustomData> data, void *frame_data, size_t len,
 
     create_kinesis_video_frame(&frame, pts, dts, flags, frame_data, len, track_id, index);
     put_frame_status = data->kinesis_video_stream->statusPutFrame(frame);
+
+    // Log fragment boundaries: keyframe = new fragment start
+    if (CHECK_FRAME_FLAG_KEY_FRAME(flags)) {
```
sirknightj, line 1275: Traced this through PIC's FrameOrderCoordinator to sanity-check the logged timestamps, since the coordinator can adjust PTS after putFrame. Good news: the logged values are safe, but two framing nits worth fixing:

1. `// Epoch ms timestamps (matching KVS backend's view)` can be misleading in AV mode. In AV mode kvssink selects FRAME_ORDERING_MODE_MULTI_TRACK_AV_COMPARE_PTS_ONE_MS_COMPENSATE_EOFR, and the coordinator buffers frames, only submitting once every track has a frame, or on EoFR ([FrameOrderCoordinator.c:319-321](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/v1.3.2/src/client/src/FrameOrderCoordinator.c#L319-L321)). When `statusPutFrame` returns, the frame generally hasn't been committed and cross-track reordering hasn't happened yet. These are submitted timestamps, not the backend/committed view - suggest rewording.
2. The logs won't show the coordinator's PTS compensation. In the AV modes the coordinator enqueues a heap copy of the frame (FrameOrderCoordinator.c:304-311) and applies its += timecodeScale overlap-compensation to that copy ([FrameOrderCoordinator.c:145](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/src/FrameOrderCoordinator.c#L145)) - never to our frame object. So `frame.presentationTs` read after `statusPutFrame` reflects the pre-submission PTS, not the adjusted one. Since fragment overlap is exactly what these logs target, worth noting they won't reveal that  1-timecode-unit nudge. (Also note that bump isn't keyframe-gated - it fires on any equal-timecode collision between different track types, and can apply more than once to a frame that keeps colliding - so the divergence isn't strictly bounded to one unit.) For VIDEO_ONLY / pass-through, none of this applies (fixupFrame is a no-op, packaging only reads PTS), so the logged value is exactly what was submitted, accurate there.

Minor: KvsSinkCustomData is shared across both tracks (gstkvssink.cpp:776), so fragment_max_pts mixes audio+video PTS in arrival order while the fragment reset only fires on video keyframes (comment to explain the behavior).

One design thought: if the goal is fragment boundaries as the backend sees them, the epoch-ms/gap logging would be more accurate in PIC layer rather than in kvssink. At that point EncodedFrameInfo gives you the authoritative post-reorder, post-timecode-compensation values. `streamState == MKV_STATE_START_CLUSTER` is the true fragment-start signal, and clusterPts/framePts are the committed timestamps. Note the existing DLOGV at [Client.c:889](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/v1.3.2/src/client/src/Client.c#L889) is at the coordinator entry (pre-buffering/reorder), so it shares the same pre-commit limitation as this kvssink log. The putFrame layer is the one that actually has the committed view. The GStreamer running-time correlation and the  first_pts/producer_start_time baseline, on the other hand, can only live in kvssink since PIC has no notion of GStreamer running time - so a split (boundaries in PIC, gst-correlation in kvssink) may be cleanest.

YPandas (approved):

PR #1300, opened 2026-07-22
kvssink: Add a new property that allows ignoring envs for region

sirknightj:
*Issue #, if available:*
- N/A (internal incident: cross-region kvssink region resolution)

*What was changed?*
- Added two opt-in boolean `kvssink` GStreamer properties, both defaulting to `false`:
  - `ignore-region-env` — ignore `AWS_DEFAULT_REGION` and always use the `aws-region` property.
  - `ignore-credentials-env` — ignore `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` and always use the `access-key` / `secret-key` / `session-token` properties.
- Added a warning log when `AWS_DEFAULT_REGION` overrides an explicitly-set `aws-region`.
- Documented both properties and the env-var precedence / region lookup order in the README section.

*Why was it changed?*
- `AWS_DEFAULT_REGION` environment variable can override the `aws-region` property. This is documented behavior (env wins by default), not an SDK defect, but consumers had no way to pin region/credentials to the pipeline configuration. These properties let a pipeline opt into ignoring the env fallback so production configuration can't be changed by earlier elements in the resolution chain.

*How was it changed?*
- Added the two properties (struct fields, enum, install, set/get, defaults) in `gstkvssink`.
- Snapshotted the AWS env vars once, up front, so resolution reads a single consistent view rather than calling `getenv` repeatedly.
- Extracted the precedence decision into pure helpers `resolveRegion` / `shouldUseCredentialsEnv` in `KvsSinkUtil` so it can be unit tested directly; defaults preserve existing env-wins behavior.

- Added table-driven unit tests (`gstkvsplugintest`, `EnvOverrideResolution` suite) covering the full region lookup matrix — env-overrides-property, property-when-env-unset, the `us-west-2` fallback when neither is set, and all `ignore-region-env` variants — plus the credential env/ignore matrix and a property round-trip test. These run without AWS credentials.
- Run the new suites from the `build` directory (built with `-DBUILD_GSTREAMER_PLUGIN=ON -DBUILD_TEST=ON`):
  ```bash
  CK_FORK=no GST_PLUGIN_PATH=$(pwd) \
    GST_CHECKS="check_region_env_override_resolution,check_credentials_env_override_resolution,check_ignore_env_properties" \
    ./tst/gstkvsplugintest
  ```
  Result: `Checks: 3, Failures: 0, Errors: 0`.
- Confirmed a clean full `make -j` build, and that `gst-inspect-1.0 kvssink` now lists both new properties:
  ```
  ignore-credentials-env: If true, ignore the AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN environment variables and always use the access-key, secret-key, and session-token properties. By default (false), these environment variables take precedence over the corresponding properties.
                        flags: readable, writable
                        Boolean. Default: false

  ignore-region-env   : If true, ignore the AWS_DEFAULT_REGION environment variable and always use the aws-region property. By default (false), the AWS_DEFAULT_REGION environment variable takes precedence over the aws-region property.
                        flags: readable, writable
                        Boolean. Default: false
  ```
- Confirmed the override warning fires when `AWS_DEFAULT_REGION` overrides an explicitly-set `aws-region` (e.g. `aws-region=us-east-1` with `AWS_DEFAULT_REGION=us-west-1`):
  ```
  export AWS_DEFAULT_REGION=us-west-1
  gst-launch-1.0 videotestsrc num-buffers=1 \
    ! x264enc ! h264parse \
    ! kvssink stream-name="my-stream" aws-region="us-east-1"
  ```

  ```
  [WARN ] [22-07-2026 05:57:43:726.052 UTC] The AWS_DEFAULT_REGION environment variable (us-west-1) is overriding the aws-region property (us-east-1) for my-stream. Set ignore-region-env=true to always use the aws-region property.
  ```

### DIFF
--- a/src/gstreamer/Util/KvsSinkResolution.cpp
+++ b/src/gstreamer/Util/KvsSinkResolution.cpp
@@ -2,18 +2,18 @@
 
 namespace kvs_sink_util {
 
+bool shouldUseEnvVar(bool ignore_env, const char *env_value) {
+    return !ignore_env && nullptr != env_value;
+}
+
 std::string resolveRegion(const std::string &property_region, const char *env_region, bool ignore_env) {
     // By default, the AWS_DEFAULT_REGION env var overrides the aws-region property. When ignore_env is
     // set, the env var is ignored and the aws-region property is always used.
     // The aws-region property has a default value of DEFAULT_REGION.
-    if (!ignore_env && nullptr != env_region) {
+    if (shouldUseEnvVar(ignore_env, env_region)) {
         return std::string(env_region);
     }
     return property_region;
-}
-
-bool shouldUseCredentialsEnv(bool ignore_env, const char *env_value) {
-    return !ignore_env && nullptr != env_value;
 }
 
 }
--- a/src/gstreamer/Util/KvsSinkResolution.cpp
+++ b/src/gstreamer/Util/KvsSinkResolution.cpp
@@ -1,0 +1,19 @@
+#include "KvsSinkResolution.h"
+
+namespace kvs_sink_util {
+
+std::string resolveRegion(const std::string &property_region, const char *env_region, bool ignore_env) {
+    // By default, the AWS_DEFAULT_REGION env var overrides the aws-region property. When ignore_env is
+    // set, the env var is ignored and the aws-region property is always used.
+    // The aws-region property has a default value of DEFAULT_REGION.
+    if (!ignore_env && nullptr != env_region) {
+        return std::string(env_region);
+    }
+    return property_region;
+}
+
+bool shouldUseCredentialsEnv(bool ignore_env, const char *env_value) {
+    return !ignore_env && nullptr != env_value;
+}
+
+}
--- a/tst/gstreamer/gstkvstest.cpp
+++ b/tst/gstreamer/gstkvstest.cpp
@@ -327,7 +327,7 @@ GST_START_TEST(check_region_env_override_resolution)
     }
 GST_END_TEST;
 
-// Exercises the credential env precedence logic directly via the pure shouldUseCredentialsEnv helper.
+// Exercises the credential env precedence logic directly via the pure shouldUseEnvVar helper.
 // Each row is one scenario; env_value == nullptr models an unset credential env var (e.g.
 // AWS_ACCESS_KEY_ID), and a non-null value models it being present. A true result means the env var
 // is consulted; false means the corresponding property wins.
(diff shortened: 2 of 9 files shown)


sirknightj:
New revision — fixes failing kvssink checks CI

* **Issue**: The kvssink unit tests job failed with exit code 9 — all 9 tests reported Received signal 11 (Segmentation fault), including pre-existing tests unrelated to this change. Assertions passed (Failures: 0); the crash was at per-test teardown.
* **Root cause**: The previous revision compiled KvsSinkUtil.cpp into the test binary so the resolution helpers could be unit tested. That TU has a LOGGER_TAG (log4cplus) static initializer plus other file-scope statics. Since KvsSinkUtil.cpp also lives in the dlopened libgstkvssink.so, those statics ended up duplicated in two modules — constructed/destructed twice — which crashed every forked test child at teardown.
* **Fix**: Extracted the two pure helpers (resolveRegion, shouldUseCredentialsEnv) into a dependency-free translation unit KvsSinkResolution.{h,cpp} (only <string>; no logging, no static initializers). The test binary now compiles that lean file instead of KvsSinkUtil.cpp, so no plugin static state is duplicated. Verified with nm that the test binary now contains only the two resolution symbols and none of KvsSinkUtil.cpp's statics. Fork isolation in the tests is unchanged.

src/gstreamer/Util/KvsSinkResolution.cpp
```diff
@@ -0,0 +1,19 @@
...
+    // The aws-region property has a default value of DEFAULT_REGION.
+    if (!ignore_env && nullptr != env_region) {
+        return std::string(env_region);
+    }
+    return property_region;
+}
+
+bool shouldUseCredentialsEnv(bool ignore_env, const char *env_value) {
+    return !ignore_env && nullptr != env_value;
+}
```
vabhasin: nit: can be made into generic `shouldUseEnvVar`?

src/gstreamer/Util/KvsSinkResolution.cpp
```diff
@@ -0,0 +1,19 @@
+#include "KvsSinkResolution.h"
+
+namespace kvs_sink_util {
+
+std::string resolveRegion(const std::string &property_region, const char *env_region, bool ignore_env) {
+    // By default, the AWS_DEFAULT_REGION env var overrides the aws-region property. When ignore_env is
+    // set, the env var is ignored and the aws-region property is always used.
+    // The aws-region property has a default value of DEFAULT_REGION.
+    if (!ignore_env && nullptr != env_region) {
```
vabhasin: can use `shouldUseEnvVar`

tst/gstreamer/gstkvstest.cpp
```diff
@@ -276,6 +277,96 @@ GST_START_TEST(check_properties_are_passed_correctly)
...
+    {
+        struct RegionCase {
+            const char *description;
+            const char *property_region; // value of the aws-region property
+            const char *env_region;      // snapshot of AWS_DEFAULT_REGION (nullptr == unset)
+            bool ignore_env;             // value of the ignore-region-env property
+            const char *expected;        // region resolveRegion should return
+        };
+
+        const RegionCase cases[] = {
```
vabhasin, line 298: Nice!

vabhasin (approved):

src/gstreamer/Util/KvsSinkResolution.cpp
sirknightj: Yes, great suggestion. I refactored the code

src/gstreamer/Util/KvsSinkResolution.cpp
sirknightj: Updated

YPandas (approved):

